PR #2, opened 2016-02-08
Python API for BFS

jkbradley:
Using conditional compilation for SQLHelpers with Spark 1.4 vs. 1.5+.  Tested 1.4 and 1.5.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@
 
 scalaVersion := "2.10.4"
 
-sparkVersion := "1.5.3-SNAPSHOT" // "1.4.1"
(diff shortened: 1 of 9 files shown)


thunterdb:
The code looks good. I am having some compilation issue to run the python tests, but it may be just my setup. I need to investigate more, feel free to merge the PR first if you want.

jkbradley:
OK thanks!  I'll go ahead and merge it then.

To run the Python tests, I needed to assemble the right version of Spark on my laptop.  Does that fix it for you?  Or better yet, is there a good way to make it download the Spark JAR from Maven instead?

PR #5, opened 2016-02-10
Python interface for the standard library

thunterdb:
It includes some basic tests to make sure the constructs are properly wrapped.

Before merging, all the examples should be consolidated with the existing `examples` package. We can finish the documentation in a separate task or do it in this task.


python/graphframes/tests.py
```diff
@@ -55,25 +55,116 @@ def setUp(self):
...
+    def test_triangle_counts(self):
+        edges = self.sqlContext.createDataFrame([(0, 1), (1, 2), (2, 0)], ["src", "dst"])
+        vertices = self.sqlContext.createDataFrame([(0,), (1,), (2,)], ["id"])
+        g = GraphFrame(vertices, edges)
+        c = g.triangle_count()
+        for row in c.vertices.select("id", "count").collect():
+            self.assertEqual(row.count, 1)
+
+
+if __name__ == "__main__":
```
thunterdb: I am going to remove this, it is useful for copy/pasting the test content in ipython.

src/main/scala/org/graphframes/GraphFramePythonAPI.scala
```diff
@@ -11,4 +11,88 @@ private[graphframes] class GraphFramePythonAPI {
...
+
+  lazy val examples: ExampleImpl = new ExampleImpl
+}
+
+/**
+ * Some standard structures in the graph literature.
+ *
+ * These are useful for running tests.
+ */
+private[graphframes] class ExampleImpl {
```
thunterdb: TODO: merge with the examples

src/main/scala/org/graphframes/lib/PageRank.scala
```diff
@@ -17,6 +17,9 @@
 
 package org.graphframes.lib
 
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types.StringType
```
thunterdb: import

thunterdb:
cc @jkbradley

jkbradley:
I'd prefer to follow up with docs, to keep PRs smaller.  I'll take a look now.

jkbradley:
It looks good.  My main comment is that I hope we can improve support to allow arbitrary Catalyst types for VertexId.  I'm seeing if I can find a good way to do that.

src/main/scala/org/graphframes/examples/TestExamples.scala
```diff
@@ -0,0 +1,126 @@
...
+      val fields = line.split(",")
+      (fields(0).toLong * 2, fields(1).toLong * 2 + 1, fields(2).toDouble)
+    }
+    val edges = sqlContext.createDataFrame(data).toDF("src", "dst", "weight")
+    val vs = data.flatMap(r => r._1 :: r._2 :: Nil).collect().distinct.map(x => Tuple1(x))
+    val vertices = sqlContext.createDataFrame(vs).toDF("id")
+    GraphFrame(vertices, edges)
+  }
+
+  private lazy val als_data =
```
thunterdb: It was easier to embed directly the file content here; the class loader could not find the resource in python tests.

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+
+    def connected_components(self):
+        """
+        Computes the connected components of the graph.
+        :return:
+        """
+        jgf = self._jvm_graph.connectedComponents().run()
+        return _from_java_gf(jgf, self._sqlContext)
+
+    def label_propagation(self, max_steps):
```
jkbradley: For these methods, shall we use the keyword_only decorator from from pyspark.ml.util (copying it in, not using the pyspark decorator)?

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+        jgf = self._jvm_graph.labelPropagation().setMaxSteps(max_steps).run()
+        return _from_java_gf(jgf, self._sqlContext)
+
+    def page_rank(self,
+                  reset_prob, source_id = None,
+                  fixed_num_iter = None,
+                  tolerance = None):
+        """
+        Runs the PageRank algorithm on the graph.
+        :param reset_prob:
```
jkbradley: set default to match Scala

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+        :param source_id: (optional) the source vertex for a personalized PageRank.
+        :param fixed_num_iter: if set, the algorithm is run for a fixed number of iterations. In this case, the
+               `tolerance` parameter is ignored.
+        :param tolerance: if set, the algorithm is run until the given tolerance. In this case, the `fixed_num_iter`
```
jkbradley: This documentation does not match the implementation.  (This raises an exception if both are set.)

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+               parameter is ignored.
+        :return:
+        """
+        builder = self._jvm_graph.pageRank().setResetProbability(reset_prob)
+        if source_id is not None:
+            builder = builder.setSourceId(source_id)
+        if fixed_num_iter is not None:
+            builder = builder.fixedIterations(fixed_num_iter)
+        if tolerance is not None:
+            assert fixed_num_iter is None
```
jkbradley: error message

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+            builder = builder.untilConvergence(tolerance)
+        jgf = builder.run()
+        return _from_java_gf(jgf, self._sqlContext)
+
+    def shortest_paths(self, landmark_ids):
```
jkbradley: fix camelCase issue here and elsewhere

python/graphframes/graphframe.py
```diff
@@ -102,6 +111,102 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+        :param num_iterations: the number of iterations to run.
+        :return:
+        """
+        jgf = self._jvm_graph.stronglyConnectedComponents().setNumIterations(num_iterations).run()
+        return _from_java_gf(jgf, self._sqlContext)
+
+    def svd_plus_plus(self, conf_dict):
+        """
+        Runs the SVD++ algorithm.
+        :param conf:
```
jkbradley: I want us to user setters, not a conf, to be consistent across the APIs.  This is why I was saying we should fix the Scala cleanups first...

python/graphframes/tests.py
```diff
@@ -77,3 +77,89 @@ def test_bfs(self):
...
+        comps = g.connected_components()
+        self.assertEqual(comps.vertices.count(), 1)
+        self.assertEqual(comps.edges.count(), 0)
+
+    def test_connected_components2(self):
+        v = self.sqlContext.createDataFrame([(0L, "a0", "b0"), (1L, "a1", "b1")], ["id", "A", "B"])
+        e = self.sqlContext.createDataFrame([(0L, 1L, "a01", "b01")], ["src", "dst", "A", "B"])
+        g = GraphFrame(v, e)
+        comps = g.connected_components()
+        self.assertEqual(comps.vertices.count(), 2)
```
jkbradley: Should test for the result of the algorithm (the new column).  Same above

python/graphframes/tests.py
```diff
@@ -77,3 +77,89 @@ def test_bfs(self):
...
+        labels2 = labels.vertices.filter("id >= 5").select("label").collect()
+        all2 = set([x.label for x in labels2])
+        self.assertEqual(all2, set([n]))
+
+    def test_page_rank(self):
+        n = 100
+        g = self._graph("star", n)
+        resetProb = 0.15
+        errorTol = 1.0e-5
+        pr = g.page_rank(resetProb, tolerance=errorTol)
```
jkbradley: test that algorithm output columns exist

python/graphframes/tests.py
```diff
@@ -77,3 +77,89 @@ def test_bfs(self):
...
+        edges = self.sqlContext.createDataFrame(all_edges, ["src", "dst"])
+        vertices = self.sqlContext.createDataFrame([(i,) for i in range(1, 7)], ["id"])
+        g = GraphFrame(vertices, edges)
+        landmarks = [1, 4]
+        g2 = g.shortest_paths(landmarks)
+        rows = g2.vertices.select("id", "distance").collect()
+
+    def test_svd_plus_plus(self):
+        g = self._graph("ALSSyntheticData")
+        (g2, cost) = g.svd_plus_plus({})
```
jkbradley: test that algorithm output columns exist

src/main/scala/org/graphframes/examples/TestExamples.scala
```diff
@@ -0,0 +1,129 @@
...
+      (i, s"node-$i")
+    }).toDF("id", "v_attr1")
+    val edges = sqlContext.createDataFrame((1 to n).map { i =>
+      (i, 0, s"edge-$i")
+    }).toDF("src", "dst", "e_attr1")
+    GraphFrame(vertices, edges)
+  }
+
+  /**
+   * Some synthetic data that sits in spark.
```
jkbradley: typo

src/main/scala/org/graphframes/lib/PageRank.scala
```diff
@@ -126,14 +128,10 @@ object PageRank {
...
-    private var srcId: Option[Long] = None
+    private var srcId: Option[Any] = None
 
-    def setSourceId[VertexId : TypeTag](srcId_ : VertexId): this.type = {
-      // TODO(tjh) This should to the conversion to the Long id.
-
-      val t = typeOf[VertexId]
-      require(typeOf[Long] =:= t, s"Only long are supported for now, type was $t")
-      srcId = Some(srcId_.asInstanceOf[Long])
+    def setSourceId[VertexId](srcId_ : VertexId): this.type = {
```
jkbradley: remove underscore from arg name

src/main/scala/org/graphframes/lib/ShortestPaths.scala
```diff
@@ -17,13 +17,14 @@
 
 package org.graphframes.lib
 
-import scala.reflect._
-import scala.reflect.runtime.universe._
+import java.util
 
-import org.apache.spark.graphx.{lib => graphxlib}
+import scala.collection.JavaConversions._
```
jkbradley: I forget why, but we always use JavaConverters instead of JavaConversions in MLlib.  I know there was a reason...

jkbradley:
Why not combine Graphs and TestExamples?

python/graphframes/graphframe.py
thunterdb: I packed all the arguments as in scala's graphX, but we can unpack all the arguments here:

```python
def svdPlusPlus(self, min_val = ..., max_val = ..., ...):
```

About using builders in python, the only place I have seen something like that being used is in GUI toolkits (wxwindows and TK bindings). The conscensus in python seems to be that the python language offers some alternatives:
http://stackoverflow.com/questions/11977279/builder-pattern-equivalent-in-python

In any case, let me know what solution you prefer.

jkbradley: * Scala: We should use a Builder which takes each parameter separately, not using a Conf.
* Python: We should use named arguments for each parameter separately, not use a Conf.

src/main/scala/org/graphframes/lib/ShortestPaths.scala
thunterdb: I learned something about scala, thanks! JavaConverters makes the conversion explicit.

thunterdb:
@jkbradley comments addressed. About separating graphs and tests examples, you are right, there is no good reason and I merged them. We still need a class and an object and a class however because python needs a proper java class to access the methods.

thunterdb:
Actually I still have to push my last change

thunterdb:
ok done

python/graphframes/graphframe.py
thunterdb: Fixed

jkbradley:
Thanks!  I'll make a final pass now

jkbradley:
Looks good.  I just had some minor cleanups, but figured a PR would be easier than comments.  Would you mind checking [https://github.com/thunterdb/graphframes-1/pull/1]?

jkbradley:
I think this is good, except that I forgot about updating the conf for SVDPlusPlus.  I'll note that as a to-do item for after we update the Scala API.

Merging now.

Thanks a lot!

PR #6, opened 2016-02-12
Belief Propagation example

jkbradley:
* Added example of BP on Ising model using conversions to GraphX + GraphX's aggregateMessages
* Made GraphXConversions.toGraphX public

PR #7, opened 2016-02-16
Aggregate messages

jkbradley:
* Wrote GraphFrame.aggregateMessages using DataFrame ops, plus some workarounds for Spark issues.  Added unit test.
* Example of BP using GraphFrame.aggregateMessages.  Added unit test
* Moved toGraphX to GraphFrame since it is very useful


jkbradley:
This depends on [https://github.com/graphframes/graphframes/pull/6] and should be rebased after [https://github.com/graphframes/graphframes/pull/6] is merged.

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -442,6 +404,13 @@ class GraphFrame protected(
   def svdPlusPlus(): SVDPlusPlus.Builder = new SVDPlusPlus.Builder(this)
 
   def triangleCount(): TriangleCount.Builder = new TriangleCount.Builder(this)
+
+  /**
+   * This is a primitive for implementing graph algorithms.
+   * This method aggregates values from the neighboring edges and vertices of each vertex.
+   * See [[AggregateMessages]] for detailed documentation.
+   */
+  def aggregateMessages: AggregateMessages.Builder = new AggregateMessages.Builder(this)
```
thunterdb: nit: to be consistent with the other algorithms, they should not take empty parens `()`, or this one should.

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -450,12 +419,15 @@ object GraphFrame extends Serializable {
   /** Column name for vertex IDs in [[GraphFrame.vertices]] */
   val ID: String = "id"
 
-  /** Column name for source vertex IDs in [[GraphFrame.edges]] */
+  /** Column name for source vertex IDs in [[GraphFrame.edges]] and [[GraphFrame.triplets]] */
```
thunterdb: If I understand correctly the triplets, this is not quite true:
 - it is the name of the column that contains the source IDs in edges
 - but it is the name of the aggregate column that contains all the vertex attributes of the source edge in the case of the triplets (this aggregate contains among other data the vertex id)

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -450,12 +419,15 @@ object GraphFrame extends Serializable {
...
   val SRC: String = "src"
 
-  /** Column name for destination vertex IDs in [[GraphFrame.edges]] */
+  /** Column name for destination vertex IDs in [[GraphFrame.edges]] and [[GraphFrame.triplets]] */
```
thunterdb: Same comment here

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -530,6 +502,49 @@ object GraphFrame extends Serializable {
...
   }
 
+  /**
+   * Given a GraphFrame and a GraphX graph derived from the GraphFrame,
+   * merge attributes from the GraphX graph into the original GraphFrame.
+   *
+   * This method is useful for doing computations using the GraphX API and then merging the results
+   * with a GraphFrame.  For example, given:
+   *  - GraphFrame `gf`
+   *  - GraphX Graph[String, Int] `gx` with a String vertex attribute we want to call "category"
```
thunterdb: you should mention somewhere that the graphX object was created from the first graphframe gf, otherwise it sounds magical and will not work for non-integral ids.

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
```diff
@@ -0,0 +1,259 @@
...
+
+    // Initialize vertex beliefs at 0.0.
+    var gx = GraphFrame(colorG.vertices.withColumn("belief", lit(0.0)), colorG.edges)
+
+    // Run BP for numIter iterations.
+    for (iter <- Range(0, numIter)) {
+      // For each color, have that color receive messages from neighbors.
+      for (color <- Range(0, numColors)) {
+        // Send messages to vertices of the current color.
+        val AM = AggregateMessages
```
thunterdb: This is to get around serialization issue of the spark context, right?

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
```diff
@@ -0,0 +1,259 @@
...
+        val aggregates = gx.aggregateMessages
+          .sendToSrc(msgForSrc)
+          .sendToDst(msgForDst)
+          .agg(sum(AM.msg).as("aggMess"))
+        val v = gx.vertices
+        // Receive messages, and update beliefs for vertices of the current color.
+        val newBeliefCol = when(v("color") === color && aggregates("aggMess").isNotNull,
+          logistic(aggregates("aggMess") + v("a")))
+          .otherwise(v("belief"))  // keep old beliefs for other colors
+        val newVertices = v
```
thunterdb: Is there any way we can do that for the user? The GraphX API returns a VertexRDD, but it seems more natural to me to return a new graph with updated vertices. This is essentially what you are doing here.

src/main/scala/org/graphframes/lib/AggregateMessages.scala
```diff
@@ -0,0 +1,161 @@
...
+     *  - aggCol: aggregate result
+     * If you need to join this with the original [[GraphFrame.vertices]], you can run an inner
+     * join of the form:
+     * {{{
+     *   val g: GraphFrame = ...
+     *   val aggResult = g.aggregateMessages.sendToSrc(msg).agg(aggFunc)
+     *   aggResult.join(g.vertices, ID)
+     * }}}
+     */
+    def agg(aggCol: Column): DataFrame = {
```
thunterdb: It took me a while to understand this portion, and I wonder what error messages will look like when users try to build complex expressions. Are there some checks we can do on the column to make sure it is going to return meaningful error messages? I am thinking of checking if this is going to be an aggregation of known attributes, etc.

src/main/scala/org/graphframes/lib/AggregateMessages.scala
```diff
@@ -0,0 +1,161 @@
...
+      require(msgToSrc.nonEmpty || msgToDst.nonEmpty, s"To run GraphFrame.aggregateMessages," +
+        s" messages must be sent to src, dst, or both.  Set using sendToSrc(), sendToDst().")
+      val triplets = g.triplets
+      import org.graphframes.GraphFrame.{DST, ID, SRC}
```
thunterdb: It is more customary to have scoped imports at the beginning of classes (you also use the same imports in other methods)

src/main/scala/org/graphframes/lib/AggregateMessages.scala
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * WARNING: This is NOT the same as [[DataFrame.cache()]].
+   *          The original DataFrame will NOT be cached.
+   *
+   * This is a workaround for SPARK-13346, which makes it difficult to use DataFrames in iterative
+   * algorithms.  This workaround converts the DataFrame to an RDD, caches the RDD, and creates
+   * a new DataFrame.  This is important for avoiding the creation of extremely complex DataFrame
+   * query plans when using DataFrames in iterative algorithms.
+   */
+  def getCachedDataFrame(df: DataFrame): DataFrame = {
```
thunterdb: You should mention eplicitly this function should not be considered a stable api

src/test/scala/org/graphframes/GraphFrameTestSparkContext.scala
```diff
@@ -31,6 +31,7 @@ trait GraphFrameTestSparkContext extends BeforeAndAfterAll { self: Suite =>
     val conf = new SparkConf()
       .setMaster("local[2]")
       .setAppName("GraphFramesUnitTest")
+      .set("spark.sql.shuffle.partitions", "4")
```
thunterdb: why do you need this change?

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
```diff
@@ -0,0 +1,259 @@
...
+        gx = GraphFrame(cachedNewVertices, gx.edges)
+      }
+    }
+
+    // Drop the "color" column from vertices
+    GraphFrame(gx.vertices.drop("color"), gx.edges)
+  }
+
+  /** More numerically stable `log(1 + exp(x))` */
+  private def log1pExp(x: Double): Double = {
```
thunterdb: there is no implementation in mllib?

thunterdb:
@jkbradley I finished to review it, it looks pretty good. I have a few comments about the API but I think they could be merged later if you want.

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
```diff
@@ -0,0 +1,259 @@
...
+
+    // Convert GraphFrame to GraphX, and initialize beliefs.
+    val gx0 = colorG.toGraphX
+    // Schema maps for extracting attributes
+    val vColsMap = colorG.vColsMap
+    val eColsMap = colorG.eColsMap
+    // Convert vertex attributes to nice case classes.
+    val gx1: Graph[VertexAttr, Row] = gx0.mapVertices { case (_, attr) =>
+      // Initialize belief at 0.0
+      VertexAttr(attr.getDouble(vColsMap("a")), 0.0, attr.getInt(vColsMap("color")))
```
thunterdb: one comment I forgot: I forgot if in the variational approximation, the coeffs should be random at start, or fixed at 0

src/main/scala/org/graphframes/GraphFrame.scala
jkbradley: I think they should not.  We still need to clean those algs up.

src/main/scala/org/graphframes/GraphFrame.scala
jkbradley: Right, thanks!

src/main/scala/org/graphframes/GraphFrame.scala
jkbradley: done

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
jkbradley: Good point.  IIRC it does not matter, though random may be smarter since BP problems are generally not convex.  Given the current problems with the randn function, I'll leave it as is.

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
jkbradley: No, it's for shorthand when referring to the src, dst, edge, and msg fields.

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
jkbradley: I asked Ankur about it, and he said there are some use cases where we do not want to union back with vertices.  It's also a little unclear what the API should look like.  Let's add it later if we think it's useful.

src/main/scala/org/graphframes/examples/BeliefPropagation.scala
jkbradley: No public one.

src/main/scala/org/graphframes/lib/AggregateMessages.scala
jkbradley: I'm not sure we'll be able to give a better error message than Spark SQL can, unless we really pick apart the user's Column to identify references.  I'm also worried about efficiency since the query analysis can get pretty hairy.  I'd prefer to leave this as is for now.  We can add checks later on if needed.

src/main/scala/org/graphframes/lib/AggregateMessages.scala
jkbradley: done

src/main/scala/org/graphframes/lib/AggregateMessages.scala
jkbradley: I think it's clear that it's a temp workaround.  As far as leaving it as a stable API, we won't need to remove it or change its behavior.  We can just deprecate it at some point.

src/test/scala/org/graphframes/GraphFrameTestSparkContext.scala
jkbradley: The default is 200.  This makes tests a lot faster on my laptop.

src/main/scala/org/graphframes/GraphFrame.scala
jkbradley: done

jkbradley:
@thunterdb Thanks for the comments!  I've updated the PR based on them.  Let me know if it's good to go.

thunterdb:
@jkbradley changes look good, feel free to merge

PR #11, opened 2016-02-20
Added GraphFrame.toString.  Fixed some doc links.

jkbradley:
Also
* fixed some doc formatting issues
* changed untilConvergence, tolerance to "tol"

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -32,7 +32,7 @@ import org.graphframes.pattern._
 
 
 /**
- * A representation of a graph using [[DataFrame]]s.
+ * A representation of a graph using `DataFrame`s.
  *
  * @groupname structure Structure information
  * @groupname conversions Conversions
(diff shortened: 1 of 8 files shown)


jkbradley:
CC @thunterdb

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -49,6 +49,16 @@ class GraphFrame private(
   /** Default constructor is provided to support serialization */
   protected def this() = this(null, null)
 
+  override def toString: String = {
+    // We call select on the vertices and edges to ensure that ID, SRC, DST always come first
+    // in the printed schema.
+    val v = vertices.select(ID, vertices.columns.filter(_ != ID) :_ *).toString
+    val e = edges.select(SRC, DST +: edges.columns.filter(c => c != SRC && c != DST) :_ *).toString
+    "GraphFrame\n" +
```
thunterdb: It would be better if the description would be a one-liner, because toString is meant to be a small description (unlike `describe`, `printSchema`, etc.) and because new lines are very annoying when `grep`ing a log file in the backend.

jkbradley: done!

thunterdb: Thanks, looks good to me!

PR #13, opened 2016-02-23
Updated user guide with code examples.  Removed GFImplicits

jkbradley:
User guide updates
* Motif finding examples for stateful motifs
* code snippets for stdlib algorithms

Code updates
* Modified motif finding tests to match user guide examples
* Removed GFImplicits


docs/user-guide.md
```diff
@@ -217,7 +259,86 @@ TODO
...
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.functions.{col, when}
+import org.graphframes.examples
+
+val g = examples.Graphs.friends  // get example graph
+
+// Query on sequence, but without state
+val chain4 = g.find("(a)-[ab]->(b); (b)-[bc]->(c); (c)-[cd]->(d)")
+val chainWith2Friends = chain4.where(
+  Seq("ab", "bc", "cd")
```
mengxr: It may be easier to understand if we expand the expression:

~~~scala
(col("ab")("relationship") === "friends").cast("int") +
  (col("bc")("relationship") === "friends").cast("int") +
  (col("cd")("relationship") === "friends").cast("int")
  >= 2)
~~~

docs/user-guide.md
```diff
@@ -217,7 +259,86 @@ TODO
...
+    .map(e => when(col(e)("relationship") === "friend", 1).otherwise(0))
+    .reduce(_ + _) >= 2)
+chainWith2Friends.show()
+
+// Query on sequence, with state (cnt)
```
mengxr: This is the same as the previous one. Do we want to call it "query with state"?

docs/user-guide.md
```diff
@@ -410,6 +541,16 @@ TODO
 
 ## Label propagation
```
mengxr: Label propagation algorithm (LPA)

It is useful to provide the wikipedia link.

docs/user-guide.md
```diff
@@ -417,7 +558,8 @@ TODO
 import org.graphframes.examples
 val g: GraphFrame = examples.Graphs.friends  // get example graph
 
-TODO
+val result = g.labelPropagation.maxSteps(5).run()
+result.vertices.select("id", "label").show()
```
mengxr: Maybe not part of this PR. It would be easy to understand the algorithm if we attach the result. Btw, `show` should be called with `truncate=false`, or users won't see the full output.

docs/user-guide.md
```diff
@@ -434,29 +577,15 @@ TODO
...
-
-TODO
-{% endhighlight %}
-</div>
+There are two implementations of PageRank.
 
-</div>
+* The first implementation uses the standalone [[GraphFrame]] interface and runs PageRank
+ for a fixed number of iterations.  This can be run by setting `numIter`.
+* The second implementation uses the `org.apache.spark.graphx.Pregel` interface and runs PageRank
```
mengxr: What happens if we set both? I think this is implementation details. Maybe it is not necessary to mention that there are two implementations in the user guide.

docs/user-guide.md
```diff
@@ -474,13 +613,25 @@ TODO
...
+results3 = g.pageRank(resetProbability=0.15, numIter=10, sourceId="a")
 {% endhighlight %}
 </div>
 
 </div>
 
-## SVD++
+## Shortest paths
+
+Computes shortest paths to the given set of landmark vertices.
```
mengxr: It would be useful to explain how to reference a landmark (by vertex ID) in the user guide.

docs/user-guide.md
jkbradley: The point of this section is to show we support stateful motifs.  I know this is more complex using sequence operations, but it is necessary for the 2nd example.  I could remove this first example, but I feel like it's kind of instructive.

docs/user-guide.md
```diff
@@ -489,7 +640,8 @@ TODO
 import org.graphframes.examples
 val g: GraphFrame = examples.Graphs.friends  // get example graph
 
-TODO
+val results = g.shortestPaths.landmarks(Seq("a", "d")).run()
```
mengxr: API-wise, it is clearer if we call it `landmarkIds`

docs/user-guide.md
mengxr: Maybe we need a better example or a precise definition for stateful motifs.

* For example, consider someone wants to plan a trip visiting several cities. He needs to spend some money between cities (on edges), and he could get some new money in cities. He wants to query all possible paths such that he won't run out of money at any location. This could be a stateful query, but I don't know how to express it clearly. It could be `var money = 10; (a where money > 0 | money += a.money)-[e where money > 0 | moeny -= e.cost]->(b where money > 0 | money += b.money)`.


* Another example is like BFS, I want to find paths that only visit each city once. So I need to mark cities as visited (state). The query might like `(a | a.state = "visited)->[e]-> (b where b.state != "visited" | b.state = "visited")`.

Just two examples, it is not very clear what we mean by "stateful motif".

docs/user-guide.md
jkbradley: it fails if we set both

jkbradley:
@mengxr I'll push an update for motif finding now.  I'll update the others later.  One thing I want to avoid is copying all API docs into the user guide.  But we can add links for each algorithm to the API docs and to Wikipedia, if we want.  At some point (not now), we will copy the GraphX user guide's relevant parts.

mengxr:
Sounds good. Feel free to merge this PR.

PR #15, opened 2016-02-24
add travis conf

mengxr:
add travis-ci conf so we can check builds against different spark versions


mengxr:
I'm merging this in so we can test other PRs easily. There is an issue with SparkContext.getOrCreate in Spark 1.6. That is why the build failed on Spark 1.6.

PR #24, opened 2016-02-24
Makefile to cross-compile the release against the different spark versions

thunterdb:
Our test jars were incorrectly built. This automates the process of creating the release jar files. It uses make instead of sbt, because I found it simpler and more reliable than trying to automate the changes with a `TaskKey`.


mengxr:
LGTM. We should consider an sbt solution, not for now, but for long-term maintenance.

PR #26, opened 2016-02-24
add a test to fail personalized page rank on Spark 1.4 and 1.5

mengxr:
I will send the fix in the next commit, which backports SPARK-11432.

This PR copies `PageRank.scala` from Spark 1.6.0 (no code changes other than the package name) and uses it for Spark 1.4 and 1.5.

@thunterdb


thunterdb:
LGTM, merging

PR #27, opened 2016-02-24
simplify triangle counts

mengxr:
If we flip the edge directions such that src < dst, the pattern `(a)-[]->(b); (b)-[]->(c); (a)-[]->(c)` would uniquely match triangles. It helps simplify the implementation. Tests are not changed in this PR.

@thunterdb


thunterdb:
nice explosion, LGTM

PR #28, opened 2016-02-24
Update BFS API to use setters for all params

mengxr:
This is more consistent with other algorithms. @thunterdb


thunterdb:
thanks, merging

PR #29, opened 2016-02-24
hide SQLHelpers and backported PageRank in API doc

mengxr:
@thunterdb


thunterdb:
thanks for the sbt magic

PR #32, opened 2016-02-24
Moves all the public exceptions to the same file

thunterdb:
cc @mengxr


mengxr:
Travis is not happy:)

jkbradley:
Why are we doing cleanups like this if they don't change the public API??

thunterdb:
The exceptions change the API (as seen in the failing test...). I also removed a block of commented out code in the process.

mengxr:
LGTM

PR #38, opened 2016-02-25
Updated Makefile and build.sbt

jkbradley:
to make packaging for spark packages easier

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ sparkVersion := sys.props.getOrElse("spark.version", "1.4.1")
 spName := "graphframes/graphframes"
 
 // Don't forget to set the version
-version := "0.1.0-SNAPSHOT"
+version := "0.1.0-spark" + sparkVersion.value.substring(0, 3)
 
 // All Spark Packages need a license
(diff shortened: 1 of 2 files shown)


jkbradley:
No hurry to merge this

mengxr:
Btw, the build doesn't work with Spark master, which uses Scala 2.11 by default. We might need to cross build this with sbt.

build.sbt
```diff
@@ -8,7 +8,7 @@ sparkVersion := sys.props.getOrElse("spark.version", "1.4.1")
 spName := "graphframes/graphframes"
 
 // Don't forget to set the version
-version := "0.1.0-SNAPSHOT"
+version := "0.1.0-spark" + sparkVersion.value.substring(0, 3)
```
mengxr: If the commit is not for the release, we should change the version back to `-SNAPSHOT`.

jkbradley: How would you recommend we specify that?

mengxr:
LGTM. I don't have good recommendations, but I can try some ideas after merge:)

PR #52, opened 2016-03-02
Minor doc cleanups post-release

jkbradley:
* Clarify directions in shortest paths
* Note that some algorithms are wrappers for GraphX and may have some scalability issues
* Add example of motif finding with negation


thunterdb:
LGTM, merging

PR #53, opened 2016-03-02
stdlib unit test updates

jkbradley:
* Added output schema checks to all unit tests for stdlib algorithms.
* Updated docs for (strongly) connected components to say component ID is of LongType.
* Moved test utilities in LabelPropagationSuite to a new TestUtils object.


thunterdb:
LGTM, merging

PR #55, opened 2016-03-07
Adding python tests to the build

thunterdb:
This PR adds the python tests to travis, and fixes a bug found in the tests.

### DIFF
--- a/src/test/resources/log4j.properties
+++ b/src/test/resources/log4j.properties
@@ -5,4 +5,5 @@ log4j.appender.console.layout=org.apache.log4j.PatternLayout
 log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
 
 log4j.logger.org.eclipse.jetty.util.component.AbstractLifeCycle=ERROR
-log4j.logger.org.apache.spark=WARN
+log4j.logger.org.apache.spark.repl.SparkIMain$exprTyper=INFO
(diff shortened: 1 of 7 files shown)


thunterdb:
@jkbradley it is finally working, please review when you have the time!

src/test/resources/log4j.properties
```diff
@@ -0,0 +1,9 @@
+log4j.rootCategory=WARN, console
+log4j.appender.console=org.apache.log4j.ConsoleAppender
+log4j.appender.console.target=System.err
+log4j.appender.console.layout=org.apache.log4j.PatternLayout
+log4j.appender.console.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n
+
+log4j.logger.org.eclipse.jetty.util.component.AbstractLifeCycle=ERROR
+log4j.logger.org.apache.spark.repl.SparkIMain$exprTyper=INFO
```
jkbradley: Is this the reason that the Python tests are logging at INFO level?  It would be nice to make them log at WARN level to be more readable (and then you could probably remove that hack above to limit the size of Travis logs).

thunterdb: Actually, this file (or any log4j file from this project) is not currently loaded by spark... I have a workaround, but it should go in a separate PR

jkbradley:
OK, I'll go ahead and merge this.  Thanks!

Issue #45, opened 2016-03-01
Sort columns from motif finding


jkbradley:
Motif finding outputs columns in an arbitrary order, but we should sort the columns to match the order of vertices and edges specified in the motif.  That way, if a user writes "(a)-[e]->(b); (b)-[e2]->(c)", then the output columns will be ordered as expected: a, e, b, e2, c.

jkbradley:
Fixed!

PR #58, opened 2016-03-10
Sort DataFrame columns from motif finding

felixcheung:
Had another attempt to reparse pattern for name, but with the need to deduplicate for pattern like `(a)-[]->(b); (b)-[]->(c); (c)-[]->(a)` which is already implemented in `seen()` thought it might be better to have a single pass at the cost of slightly less readable code with pair as return values.

Thought? #45

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -224,7 +224,8 @@ class GraphFrame private(
    * For example, `graph.find("(a)-[e]->(b); (b)-[e2]->(a)")` will search for pairs of vertices
    * `a,b` connected by edges in both directions.  It will return a `DataFrame` of all such
    * structures in the graph, with columns for each of the named elements (vertices or edges)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -367,12 +369,13 @@ class GraphFrame private(
   private def findSimple(
       prevPatterns: Seq[Pattern],
       prevDF: Option[DataFrame],
-      remainingPatterns: Seq[Pattern]): DataFrame = {
+      prevNames: Seq[String],
+      remainingPatterns: Seq[Pattern]): (DataFrame, Seq[String]) = {
```
jkbradley: Document new argument & return value

jkbradley:
That's really the only issue I see. Otherwise, looks good.  Thanks!

felixcheung:
done, thanks!

jkbradley:
LGTM
Thanks for the fix!

PR #66, opened 2016-03-18
motif finding should avoid name conflict when matching with anonymous edge

### LINKED ISSUE #65: motif finding should avoid name conflict when matching with anonymous edge
Currently it is using `__tmp` as column name which could theoretically conflict with an actual column with that name.
See #50

felixcheung:
Fix #65 with randomly generated name
@jkbradley

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -17,9 +17,9 @@
 
 package org.graphframes
 
-import org.apache.log4j.PropertyConfigurator
+import java.util.Random
 
-import scala.reflect.runtime.universe.TypeTag
+import org.apache.log4j.PropertyConfigurator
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -32,6 +32,8 @@ import org.apache.spark.storage.StorageLevel
 import org.graphframes.lib._
 import org.graphframes.pattern._
 
+import scala.reflect.runtime.universe.TypeTag
```
jkbradley: Let's follow Spark style: This should go between the Java & the org.* imports.

jkbradley:
This looks good, but could you please add a little unit test which fails before your fix?  Thanks!

felixcheung:
Done. Thanks for the feedback!

jkbradley:
LGTM, thanks again!

PR #75, opened 2016-04-04
Updated TriangleCount doc to mention update to GraphX 2.0

jkbradley:
GraphX 2.0 includes a fix for SPARK-3650, making GraphX's triangleCount behave like GraphFrames by default.  This doc update mentions this change.


jkbradley:
CC: @felixcheung Would you mind taking a look (to confirm that this is correct)?  Thanks!

felixcheung:
looks good (as per https://github.com/apache/spark/pull/11290)

felixcheung:
should it just call graphx triangleCount if running with Spark 2.0?

jkbradley:
I haven't done benchmarking, so it's hard to say.  I'll make an issue for it.  I bet GraphX would be faster, but DataFrames would be more scalable (and eventually faster once we port some optimizations).

jkbradley:
Thanks for taking a look!  Merging

Issue #64, opened 2016-03-16
Add cache(), persist() methods


jkbradley:
They should just call cache() and persist() on both DataFrames.

felixcheung:
btw, I'm having real problems with .cache() calls on DataFrames related to the graph.
This is what I have, simplified:
```
...
g = GraphFrame(v, e)
m = g.find("(a)-[]->(b)")
m.cache()
m.show()
```

This is probably unrelated to this issue specifically, but I find that if I have this `m.cache()` calls, it takes 2x-3x time for csv-backed edge/vertice DataFrames, and for Parquet-backed edge/vertice DataFrame, it just hangs.
I don't have a simple repro but I can try to recreate it.

jkbradley:
That sounds bad.  It must be a DataFrames issue since GraphFrames isn't doing anything fancy (just joins).  A reproduction of it would be great, if possible, thanks!

thunterdb:
Closing this ticket.

PR #78, opened 2016-04-13
Add cache(), persist() methods

felixcheung:
See #64
Add API in Scala and Python

I couldn't find a way to test this - checks for queryExecution can't access some internal field which is needed to check if the operation is cached.

### DIFF
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -16,8 +16,8 @@
 #
 
 from pyspark import SparkContext
-from pyspark.storagelevel import StorageLevel
 from pyspark.sql import DataFrame, SQLContext
+from pyspark.storagelevel import StorageLevel
 
 def _from_java_gf(jgf, sqlContext):
     """
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -16,6 +16,7 @@
 #
 
 from pyspark import SparkContext
+from pyspark.storagelevel import StorageLevel
 from pyspark.sql import DataFrame, SQLContext
 
 def _from_java_gf(jgf, sqlContext):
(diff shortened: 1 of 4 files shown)


python/graphframes/graphframe.py
```diff
@@ -16,6 +16,7 @@
 #
 
 from pyspark import SparkContext
+from pyspark.storagelevel import StorageLevel
```
jkbradley: sort imports

python/graphframes/graphframe.py
```diff
@@ -96,6 +97,28 @@ def edges(self):
...
+
+    def persist(self, storageLevel=StorageLevel.MEMORY_ONLY):
+        """Persist the dataframe representation of vertices and edges of the graph with the given
+        storage level.
+        """
+        javaStorageLevel = self._sc._getJavaStorageLevel(storageLevel)
+        self._jvm_graph.persist(javaStorageLevel)
+        return self
+
+    def unpersist(self, blocking=True):
```
jkbradley: Set default blocking to False to match DataFrame/Dataset

src/test/scala/org/graphframes/GraphFrameSuite.scala
```diff
@@ -223,4 +224,13 @@ class GraphFrameSuite extends SparkFunSuite with GraphFrameTestSparkContext {
...
     assert(degrees === Map(1L -> 2, 2L -> 3, 3L -> 1))
   }
+
+  test("cache") {
+    val g = GraphFrame(vertices, edges)
+
+    g.persist(StorageLevel.MEMORY_ONLY)
+
+    g.unpersist()
+    // org.apache.spark.sql.execution.columnar.InMemoryRelation is private and not accessible
```
jkbradley: State that this prevents us from easily verifying whether it is cached.

jkbradley:
Thanks for the PR!  I agree it'd be hard AFAIK to test for caching without using the internal APIs.  I think this is fine for now.

python/graphframes/graphframe.py
felixcheung: I think the default is True?
[Python API doc](http://spark.apache.org/docs/latest/api/python/pyspark.sql.html?highlight=unpersist#pyspark.sql.DataFrame.unpersist)
[Python code](https://github.com/apache/spark/blob/master/python/pyspark/sql/dataframe.py#L329)

python/graphframes/graphframe.py
felixcheung: I thought this is sorted?
```
from pyspark import SparkContext
from pyspark.storagelevel import StorageLevel
from pyspark.sql import DataFrame, SQLContext
```
`storagelevel` < `sql`
? ;)

python/graphframes/graphframe.py
jkbradley: Oh!  I was looking at Scala.  I'll ping others about fixing this inconsistency in Spark 2.0: [https://issues.apache.org/jira/browse/SPARK-14717]

felixcheung: Would you like me to change GF to default to False?

jkbradley:
LGTM
Thanks @felixcheung !
Merging with master

akshayagarwalgithub:
Hi Team @jkbradley @felixcheung . I just wanted to be on same page on this. Can we cache a graph in memory or we can not?

felixcheung:
Yes - is there a problem you are running into?

akshayagarwalgithub:
Hi @felixcheung and @jkbradley  though I could cache a graphframe but did'nt see any performance improvement when I ran next algorithm on that graphframe. 
Is there a different approach I need to take?

PR #88, opened 2016-06-02
Removed explicit typing of long integers in tests.py

fred-lefebvre:
Removed explicit typing of integers to long to make code compatible with Python3

The use of long integers doesn't seem required here anyway.  If it was, an alternative would be to check the python version before declaring those array and adapt typing accordingly.


thunterdb:
Merging, thanks for the pull request!

PR #90, opened 2016-06-09
Update user-guide.md

qingpeng:
add SQL package import in example code

I got error with the next line in my experiment.

scala> val vertexInDegrees: DataFrame = g.inDegrees
<console>:31: error: not found: type DataFrame
         val vertexInDegrees: DataFrame = g.inDegrees
                              ^

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -146,6 +146,9 @@ g.edges.show()
 // |  a|  e|      friend|
 // +---+---+------------+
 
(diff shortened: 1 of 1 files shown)


rxin:
Can you just import DataFrame rather than everything?

qingpeng:
Sure. Just changed that.
Import DataFrame only now

rxin:
Thanks - merging in.

PR #97, opened 2016-07-19
Java example project

kamir:
I started an embedded sub-project (Maven based) for Java examples as a first step towards a user guide for Java users interested in graphframes.


kamir:
Hello GraphFrames team,
what is the next thing I can do in order to get the contribution merged?
Just started work on a graph generator which produces GraphFrames.

mengxr:
@kamir Thanks for the PR. But it might not be the right time to add Java examples at this stage:

1. The APIs are experimental, which might change in the future. Adding more example code means more maintenance work in the future.
2. We don't have any Java compatibility tests yet. To enhance Java support, the right order would be: a) add Java compatibility unit tests, b) add Java code snippets to the user guide (http://graphframes.github.io/user-guide.html), c) add full examples (probably after Scala/Python).

So I'm closing this PR for now. Sorry but thanks again for writing examples in Java!

PR #115, opened 2016-10-17
use monotonically_increasing_id in zipWithUniqueId for 2.x

mengxr:
Both SPARK-9020 and SPARK-13473 were fixed in 2.x. This PR removes the workaround to get unique long IDs in 2.x.

### DIFF
--- a/src/main/spark-2.x/org/apache/spark/sql/SQLHelpers.scala
+++ b/src/main/spark-2.x/org/apache/spark/sql/SQLHelpers.scala
@@ -18,8 +18,8 @@
 package org.apache.spark.sql
 
 import org.apache.spark.sql.catalyst.expressions.Expression
-import org.apache.spark.sql.functions.udf
-import org.apache.spark.sql.types.{DataType, LongType, StructField, StructType}
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.DataType
 
 object SQLHelpers {
   def getExpr(col: Column): Expression = col.expr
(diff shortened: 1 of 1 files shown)


src/main/spark-2.x/org/apache/spark/sql/SQLHelpers.scala
```diff
@@ -28,17 +28,11 @@ object SQLHelpers {
...
-    val sqlContext = df.sqlContext
-    val schema = df.schema
-    val rdd = df.rdd.zipWithUniqueId().map { case (row, id) =>
-      Row(row, id)
-    }
-    val outputSchema = StructType(Seq(
-      StructField("row", schema, false), StructField("uniq_id", LongType, false)))
-    sqlContext.createDataFrame(rdd, outputSchema)
+    df.select(
+      struct(df.columns.map(c => df(c)) :_*).as("row"),
```
jkbradley: I'm encouraging people to use ```nestAsCol```.

jkbradley:
LGTM
I had one comment, if you don't mind fixing it.  Thanks!

jkbradley:
Oh nevermind, that looks like an actual test failure

src/main/spark-2.x/org/apache/spark/sql/SQLHelpers.scala
mengxr: That method is not visible under this package.

mengxr:
Closing this PR since `monotonically_increasing_id` still has problems. https://issues.apache.org/jira/browse/SPARK-14393

Issue #117, opened 2016-10-18
More scalable connected components implementation


jkbradley:
There have been many reports of the connected components algorithm in GraphX and GraphFrames not scaling.  @mengxr has a prototype of a better algorithm.  This issue is for tracking adding it to GraphFrames master.

Subtasks:
- [x] Implementation: #119
- [x] Improved unit tests: #121
- [x] GraphX legacy support: #122
- [x] Python API: #123
- [x] (optional) checkpoint interval param: #124
- [x] handle skewness in assigning long IDs

nchammas:
Just curious: Is this related at all to the issue described in #116?

jkbradley:
I'm pretty sure not.

jkbradley:
https://github.com/graphframes/graphframes/pull/119

mengxr:
Added one more TODO to handle skewness in assigning long IDs. Otherwise, CC might fail at the first step on skewed graphs.

PR #119, opened 2016-11-01
New connected components implementation

mengxr:
This PR replaces the GraphX implementation of connected components by a DataFrame-based implementation. It also switched the algorithm to the one proposed in "Connected Components in MapReduce and Beyond". The new implementation handles skewness by separating the joins into broadcast join of vertices of high degrees and normal join of the rest.

The implementation was tested on a graph w/ ~2 billion edges, but without long ID conversion.

Note that it uses `indexedVertices` and `indexedEdges` to assign long IDs. This might have performance issues if the input graph is skewed. We can do a follow-up PR to address it.

The unit tests are updated to check the broadcast join path. The coverage is apparently not ideal at this time.

cc: @jkbradley

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -37,6 +37,9 @@ import org.graphframes.{GraphFrame, Logging}
  *
  * The computation is done using alternating large star and small star iterations proposed in
  * "Connected Components in MapReduce and Beyond" with skewed join optimization.
+ * Intermediate data is checkpointed under [[org.apache.spark.SparkContext.getCheckpointDir]]
+ * with prefix "connected-components".
+ * If the checkpoint directory is not set, this throws nn [[IOException]].
  *
  * The resulting DataFrame contains all the vertex information and one additional column:
  *  - component (`LongType`): unique ID for this component
@@ -51,22 +54,23 @@ class ConnectedComponents private[graphframes] (
   private var broadcastThreshold: Int = 1000000
 
   /**
-    * Sets broadcast threshold in propagating component assignments (default: 1000000).
-    * If a node degree is greater than this threshold at some iteration, its component assignment
-    * will be collected and then broadcasted back to propagate the assignment to its neighbors.
-    * Otherwise, the assignment propagation is done by a normal Spark join.
+   * Sets broadcast threshold in propagating component assignments (default: 1000000).
+   * If a node degree is greater than this threshold at some iteration, its component assignment
+   * will be collected and then broadcasted back to propagate the assignment to its neighbors.
+   * Otherwise, the assignment propagation is done by a normal Spark join.
    */
   def setBroadcastThreshold(value: Int): this.type = {
+    require(value >= 0, s"Broadcast threshold must be non-negative but got ${value}.")
     broadcastThreshold = value
     this
   }
 
   /**
-    * Gets broadcast threshold in propagating component assignment.
-    *
-    * @see [[setBroadcastThreshold()]]
-    */
-  def getBroadcastThreshold: Long = broadcastThreshold
+   * Gets broadcast threshold in propagating component assignment.
+   *
+   * @see [[setBroadcastThreshold()]]
+   */
+  def getBroadcastThreshold: Int = broadcastThreshold
 
   /**
    * Runs the algorithm.
@@ -106,10 +110,10 @@ private object ConnectedComponents extends Logging {
    * Prepares the input graph for computing connected components by:
    *   - de-duplicating vertices and assigning unique long IDs to each,
    *   - changing edge directions to have increasing long IDs from src to dst,
-   *   - de-duplicating edges and removing loops.
+   *   - de-duplicating edges and removing self-loops.
    * In the returned GraphFrame, the vertex DataFrame has two columns:
    *   - column `id` stores a long ID assigned to the vertex,
-   *   - column `orig_id` stores the original ID.
+   *   - column `attr` stores the original vertex attributes.
    * The edge DataFrame has two columns:
    *   - column `src` stores the long ID of the source vertex,
    *   - column `dst` stores the long ID of the destination vertex,
@@ -205,7 +209,7 @@ private object ConnectedComponents extends Logging {
       throw new IOException(
         "Checkpoint directory is not set. Please set it first using sc.setCheckpointDir")
     }
-    logger.info(s"$logPrefix Use $checkpointDir for checkpointing.")
+    logger.info(s"$logPrefix Using $checkpointDir for checkpointing.")
 
     logger.info(s"$logPrefix Preparing the graph for connected component computation ...")
     val g = prepare(graph)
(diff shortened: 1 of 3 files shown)


mengxr:
This PR depends on #120. Otherwise, it will fail in 1.4 and 1.5 builds.

nchammas:
Couple of quick questions:

1. Will PySpark automatically benefit from this improved implementation, or is separate work required for that?
2. Is a notable performance difference expected with this new implementation?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+    var prevSum = Long.MaxValue
+    while (!converged) {
+      val toBeCleaned = mutable.ArrayBuffer.empty[DataFrame]
+      toBeCleaned += ee
+      // compute min neighbors including self
+      val minNbrs1 = minNbrs(ee)
+        .withColumn(MIN_NBR, minValue(col(SRC), col(MIN_NBR)).as(MIN_NBR))
+        .persist(StorageLevel.MEMORY_AND_DISK)
+      toBeCleaned += minNbrs1
+      // connect all strictly larger neighbors to the min neighbor (including self)
```
jkbradley: Note that this is the large-star op

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+      ee = skewedJoin(ee, minNbrs1, broadcastThreshold, logPrefix)
+        .select(col(DST).as(SRC), col(MIN_NBR).as(DST)) // src > dst
+        .distinct()
+        .persist(StorageLevel.MEMORY_AND_DISK)
+      toBeCleaned += ee
+      // compute min neighbors
+      val minNbrs2 = minNbrs(ee)
+        .persist(StorageLevel.MEMORY_AND_DISK)
+      toBeCleaned += minNbrs2
+      // connect all smaller neighbors to the min neighbor
```
jkbradley: Note that this is the small-star op

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+    val vv = g.vertices
+    var ee = g.edges.persist(StorageLevel.MEMORY_AND_DISK)
+    val numEdges = ee.count()
+    logger.info(s"$logPrefix Found $numEdges edges after preparation.")
+
+    var converged = false
+    var k = 0
+    var prevSum = Long.MaxValue
+    while (!converged) {
+      val toBeCleaned = mutable.ArrayBuffer.empty[DataFrame]
```
jkbradley: toBeCleaned -> toBeUnpersisted

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+        .distinct()
+      // TODO: remove this after DataFrame.checkpoint is implemented
+      val out = s"$checkpointDir/$k"
+      ee.write.parquet(out)
+      if (k > 0) {
+        FileSystem.get(sc.hadoopConfiguration)
+          .delete(new Path(s"$checkpointDir/${k - 1}"), true)
+      }
+      toBeCleaned.foreach(df => df.unpersist(true))
+      System.gc() // hint Spark to clean shuffle directories
```
jkbradley: Map GC hint across workers too?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+ *
  * The resulting DataFrame contains all the vertex information and one additional column:
  *  - component (`LongType`): unique ID for this component
  *
  * The resulting edges DataFrame is the same as the original edges DataFrame.
+ *
+ * @see https://mmds-data.org/presentations/2014/vassilvitskii_mmds14.pdf
  */
-class ConnectedComponents private[graphframes] (private val graph: GraphFrame) extends Arguments {
+class ConnectedComponents private[graphframes] (
```
mengxr: done

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+      .select(SRC, MIN_NBR)
+      .as[(Long, Long)]
+      .collect()
+      .toMap // TODO: use OpenHashMap
+    if (hubs.isEmpty) {
+      return edges.join(minNbrs, SRC)
+    } else {
+      logger.debug(s"$logPrefix Number of skewed keys: ${hubs.size}.")
+    }
+    val isHub = udf { id: Long =>
```
mengxr: I don't quite understand this comment.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
+    hashJoined.unionAll(broadcastJoined)
+  }
+
+  /**
+   * Runs connected components with default parameters.
+   */
   def run(graph: GraphFrame): DataFrame = {
-    val gx = graphxlib.ConnectedComponents.run(graph.cachedTopologyGraphX)
-    GraphXConversions.fromGraphX(graph, gx, vertexNames = Seq(COMPONENT_ID)).vertices
+    new ConnectedComponents(graph).run()
```
mengxr: Good point. Shall I make another PR for it?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,262 @@
...
   }
 
-  private val COMPONENT_ID = "component"
+  private def run(graph: GraphFrame, broadcastThreshold: Int): DataFrame = {
+    val runId = UUID.randomUUID().toString.takeRight(8)
+    val logPrefix = s"[CC $runId]"
+    logger.info(s"$logPrefix Start connected components with run ID $runId.")
```
mengxr: `logger.info` uses macro so it doesn't unnecessarily materialize the message. I'm not sure whether `logInfo` behaves the same.

mengxr:
@nchammas The python interface should use the new implementation automatically, though I haven't update it to take `broadcastThreshold` as an arg.

The new implementation pays more attention to scalability issues we observed on huge graphs. For smaller graphs, it could be slower due to skewed joins and checkpointing. We are working on several possible optimizations, including:

1) DataFrame checkpointing
2) support checkpoint interval
3) better/reliable long ID assignment

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -17,38 +17,275 @@
...
  * Connected components algorithm.
  *
  * Computes the connected component membership of each vertex and returns a DataFrame of vertex
  * information with each vertex assigned a component ID.
  *
+ * The computation is done using alternating large star and small star iterations proposed in
+ * "Connected Components in MapReduce and Beyond" with skewed join optimization.
+ * Intermediate data is checkpointed under [[org.apache.spark.SparkContext.getCheckpointDir]]
+ * with prefix "connected-components".
+ * If the checkpoint directory is not set, this throws nn [[IOException]].
```
jkbradley: nn -> an

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
jkbradley: (discussed offline, here for the record:)  We could leave the hubs in a DataFrame instead of collecting them as a map.  Then the broadcast join could be done using a broadcast hint.  To be done eventually once we believe it is more efficient.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
jkbradley: logInfo calls logger.info, just as in Spark.  not important though

jkbradley:
Thanks @mengxr !  Merging now

jkbradley:
Linking with issue: https://github.com/graphframes/graphframes/issues/117

Issue #122, opened 2016-11-03
Connected Components: support legacy GraphX implementation


jkbradley:
Add option to use GraphX instead of new CC implementation

Issue #121, opened 2016-11-03
Connected Components: improve unit tests


jkbradley:
We should improve unit test coverage for CC before we release the new version.

mavencode01:
@mengxr I'm experiencing an issue with  connected component, basically I'm trying to cluster a hierarchical dataset. It works fine for small dataset, I could clustered the data into separate clusters. 

However, for large hierarchical dataset (about `1.60million vertices`) the result seems wrong. 
The resulting clusters from connected component have many intersections. This should not be the case. I expect the hierarchical dataset to be clustered into separate smaller clusters.

    ...
    val vertices = universe.map(u => (u.id, u.username, u.age, u.gamescore))
                              .toDF("id", "username", "age","gamescore")
                              .alias("v")


    val lookup = sparkSession.sparkContext.broadcast(universe.rdd.collectAsMap())


    def buildEdges(src: String, dest: String) = {
        Edge(lookup.value.get(src).get, lookup.value.get(dest).get, 0)
    }


    val edges  =  similarityDatasetNoJboss.mapPartitions(_.map(s => buildEdges(s.username1, s.username2)))
                                          .toDF("src", "dst", "default")

    val graph = GraphFrame(vertices, edges)

    val cc = graph.connectedComponents.run().select("id", "component")


Do some validation test

    Select id, count(component)
    group by id

I expect each id to belong to one cluster/component and count = 1

jkbradley:
@mavencode01 Do you know if your ```vertices``` DataFrame has multiple rows with the same ID?  If so, that might cause the bug you're seeing.  GraphFrames assumes that the vertices DataFrame is valid and does not verify or dedup.

mengxr:
@mavencode01 please create a new issue to continue the discussion. This one is only for adding new unit tests. I'm going to mark it as closed. Thanks!

PR #125, opened 2016-11-03
[#122] Add option to use GraphX for connected components

mengxr:
This PR adds a new parameter `algorithm` to `ConnectedComponents`, so users can use the implementation in GraphX by setting it to `"graphx"`. The default algorithm is called "graphframes". I thought about giving the algorithm a better name. But the names are either too long, e.g., "alternating-large-star-small-star", or hard to remember, e.g., "kiveris-2014".

Unit tests for `friends` graph are updated to test both code paths. More unit tests will be added as part of #121.

cc: @jkbradley


mengxr:
It seems that `collect_list` was defined but not registered in 1.6.2.

mengxr:
@jkbradley This should be ready for review.

jkbradley:
LGTM
Merging
Thanks!

Issue #124, opened 2016-11-03
Connected Components: checkpoint interval param


jkbradley:
Add configurable checkpoint interval for new CC implementation

felixcheung:
PR #126

jkbradley:
Fixed!

PR #126, opened 2016-11-05
[#124] Support checkpoint interval in connected components

mengxr:
This PR adds `checkpointInterval` to the new CC implementation. The logic for removing out-of-scope persisted DataFrames was removed to simplify the code. Those should be cleaned by Spark automatically. `System.gc()` gives hint.

The default value is 1 to avoid exploding the query plan. This might change after we support query plan checkpointing in Spark SQL.

cc: @jkbradley


jkbradley:
LGTM
Merging with master
Thanks!

PR #127, opened 2016-11-07
add empty and chain graphs to Graphs

mengxr:
This PR adds two simple graphs to `examples.Graphs`:

* empty graph with arbitrary ID type (that can be recognized by SQL)
* chain graph

The purpose is to simplify unit tests coming in #121.


mengxr:
merged since this is a trivial change

PR #128, opened 2016-11-08
use DecimalType for the sum in connected components

mengxr:
This PR uses decimal type to compute the sum to avoid possible overflow issues. Note that even with long integer overflow, it is very unlikely to have `prevSum == currSum` happen before convergence. This PR just removed that possibility.

cc: @jkbradley

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -343,13 +343,19 @@ private object ConnectedComponents extends Logging {
 
       // Taking the sum in DecimalType to preserve precision.
       // We use 20 digits for long values and Spark SQL will add 10 digits for the sum.
-      // It should be able to handle 100 billion edges without overflow.
+      // It should be able to handle 200 billion edges without overflow.
       val (currSum, cnt) = ee.select(sum(col(SRC).cast(DecimalType(20, 0))), count("*")).rdd
         .map { r =>
           (r.getAs[BigDecimal](0), r.getLong(1))
         }.first()
       if (cnt != 0L && currSum == null) {
-        throw new ArithmeticException("The total sum of current assignments exceeded limit 1e30.")
+        throw new ArithmeticException(
+          s"""
+             |The total sum of edge src IDs is used to determine convergence during iterations.
+             |However, the total sum at iteration $iteration exceeded 30 digits (1e30),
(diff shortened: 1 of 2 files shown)


src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -346,9 +340,14 @@ private object ConnectedComponents extends Logging {
       ee.persist(StorageLevel.MEMORY_AND_DISK)
 
       // test convergence
-      val currSum = ee.select(sum(col(SRC))).as[Long].first()
+
+      // Taking the sum in DecimalType to preserve precision.
+      // We use 20 digits for long values and Spark SQL will add 10 digits for the sum.
+      // It should be able to handle 100 billion edges without overflow.
```
jkbradley: What happens with DecimalType overflow?  Will the value wrap or become null?

mengxr: `null`

jkbradley: Ok then I think we should either document that this can fail silently for > 100 billion edges or explicitly check and throw an exception.

mengxr: Added a check.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -346,9 +340,20 @@ private object ConnectedComponents extends Logging {
...
+      val (currSum, cnt) = ee.select(sum(col(SRC).cast(DecimalType(20, 0))), count("*")).rdd
+        .map { r =>
+          (r.getAs[BigDecimal](0), r.getLong(1))
+        }.first()
+      if (cnt != 0L && currSum == null) {
+        throw new ArithmeticException("The total sum of current assignments exceeded limit 1e30.")
```
jkbradley: I don't think users will understand this.  How about "The number of edges in this graph exceed 1e30, but Connected Components cannot support more than 1e30 edges."?

mengxr: 1e30 is not the number of edges. It is the sum of component IDs assigned. The component IDs could be as big as 1e19 (Long.MaxValue). So we can handle ~200 billion edges. I will update the message to make the message clearer.

jkbradley:
LGTM
merging
Thanks!

PR #129, opened 2016-11-08
[#121] add more unit tests to connected components

mengxr:
This PR adds more unit tests to connected components, mostly verifying the results on simple graph topologies. It does reveal two bugs:

1. I forgot to set the default value of `algorithm` param.
2. `indexedVertices` doesn't always cast ID to long.

cc: @jkbradley

### DIFF
--- a/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
+++ b/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
@@ -95,10 +95,10 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
   }
 
   test("star graph") {
-    val n = 5
-    val g = Graphs.star(5)
+    val n = 5L
+    val g = Graphs.star(5L)
     val components = g.connectedComponents.run()
-    val expected = Set((0 to n).toSet) // star graph uses Int IDs instead of Long
+    val expected = Set((0L to n).toSet)
     assertComponents(components, expected)
   }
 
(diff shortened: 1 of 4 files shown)


src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
```diff
@@ -105,20 +181,21 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
...
-    val actualComponents = actual.select("component", "id").rdd
-      .map { case Row(component: Long, id: T) =>
-        (component, id)
-      }.groupByKey()
+  private def assertComponents[T: ClassTag:TypeTag](
+      actual: DataFrame,
+      expected: Set[Set[T]]): Unit = {
+    import actual.sqlContext.implicits._
+    // note: not using agg + collect_list because collect_list is not available in 1.6.2 w/o hive
+    val actualComponents = actual.select("component", "id").as[(Long, T)].rdd
```
mengxr: Switched to `as` because in Spark 1.6 it returns `java.lang.Long` instead of `long`, which messed up case matching.

src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
```diff
@@ -59,8 +86,56 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
...
+    val components = g.connectedComponents.run()
+    val expected = Set((0L until n).toSet)
+    assertComponents(components, expected)
+  }
+
+  test("star graph") {
+    val n = 5
+    val g = Graphs.star(5)
+    val components = g.connectedComponents.run()
+    val expected = Set((0 to n).toSet) // star graph uses Int IDs instead of Long
```
jkbradley: Shall we just change them to Long?  I doubt anyone will notice the API change.

thunterdb:
Please do not merge before #132  is merged first. The tests are currently failing.

jkbradley:
LGTM pending tests

Issue #123, opened 2016-11-03
Connected Components: Python API for new implementation


jkbradley:
Need to add params to fully support new implementation

PR #130, opened 2016-11-08
[#123] update connected component python interface

mengxr:
Updated the Python interface of connected components to expose params from the new implementation (pending travis).

### DIFF
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -224,7 +224,7 @@ class GraphFrame(object):
         """
         jdf = self._jvm_graph.connectedComponents() \
             .setAlgorithm(algorithm) \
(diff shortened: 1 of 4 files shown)


python/graphframes/graphframe.py
```diff
@@ -207,15 +207,26 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
 
     # Standard algorithms
 
-    def connectedComponents(self):
+    def connectedComponents(self, algorithm = "graphframes", checkpointInterval = 1,
+                            broadcastThreshold = 1000000):
```
thunterdb: if the threshold is too large, it may become a python `long` and fail to get passed to the jvm. This is probably not going to happen anytime soon.

PR #132, opened 2016-11-09
Fixes the connected components tests

thunterdb:
They were failing because the checkpoint directory was not explicitly set. Also, the error would not get reported because of bash 'features'.

### DIFF
--- a/python/graphframes/tests.py
+++ b/python/graphframes/tests.py
@@ -16,6 +16,8 @@
 #
 
 import sys
+import tempfile
+import shutil
 
 if sys.version_info[:2] <= (2, 6):
     try:
@@ -37,7 +39,8 @@ class GraphFrameTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.sc = SparkContext('local[4]', cls.__name__)
-        cls.sc.setCheckpointDir("/tmp")
+        cls.checkpointDir = tempfile.mkdtemp()
+        cls.sc.setCheckpointDir(cls.checkpointDir)
         cls.sql = SQLContext(cls.sc)
         # Small tests run much faster with spark.sql.shuffle.partitions=4
         cls.sql.setConf("spark.sql.shuffle.partitions", "4")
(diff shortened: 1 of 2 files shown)


python/run-tests.sh
```diff
@@ -56,10 +56,13 @@ set -e
 # Horrible hack for spark 1.4: we manually remove some log lines to stay below the 4MB log limit on Travis.
 # To remove when we ditch spark 1.4.
```
jkbradley: We can remove this now!

python/run-tests.sh
```diff
@@ -55,11 +55,14 @@ set -e
...
+# Since we pipe to remove the output, we need to use some horrible BASH features:
+# http://stackoverflow.com/questions/1221833/bash-pipe-output-and-capture-exit-status
+test ${PIPESTATUS[0]} -eq 0 || exit 1
 
 # Run doc tests
 
 cd "$DIR"
 
-exec python -u ./graphframes/graphframe.py "$@"
+python -u ./graphframes/graphframe.py "$@"
```
jkbradley: Not sure myself: Why did we have exec before, and why does it matter to have or not have it?

thunterdb: we do not need it here, I think, because we do not need to fork commands in the background.

python/run-tests.sh
thunterdb: Unfortunately, I see a lot of logs being generated for the spark 1.6 run. The spark 2 travis run does not print logs, as expected from the log4j file.

python/graphframes/tests.py
```diff
@@ -37,6 +37,7 @@ class GraphFrameTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.sc = SparkContext('local[4]', cls.__name__)
+        cls.sc.setCheckpointDir("/tmp")
```
thunterdb: Fixed

jkbradley:
LGTM pending tests
Thanks a lot for noticing & fixing this!

thunterdb:
@jkbradley thanks, merging

Issue #85, opened 2016-05-23
How to use is with python3?


renewjoy:
When I runt this import with pyspark with python3.51, it errors:

> from graphframes import *
> ZipImportError: can't find module 'graphframes'

but is success in pyspark with python 2.7.1

mortada:
I'm running into the same issue

thunterdb:
The current infrastructure of spark-packages publishes python bytecode (`.pyc` files) instead of source files. Since the package was prepared with python2, I believe this is the issue.

Unless you have some specific ways to address this issue (i.e. a pull request), I am going to wait for the next release of sbt-spark-packages which should alleviate the bytecode issue.

nchammas:
@thunterdb - Can you expand on how this issue will be addressed by `sbt-spark-packages`? I'm curious.

By the way, the modern convention in Python for releasing artifacts is to publish [wheels](https://packaging.python.org/distributing/#wheels). Are there plans to use them?

psinger:
Is there any news regarding this? Makes it impossible for me to use the package currently as Im running everything with python 3.

nchammas:
@thunterdb - The latest [release of `sbt-spark-packages`](https://github.com/databricks/sbt-spark-package/releases) appears to be from August 15, a month after your latest comment. Is that release supposed to help fix this issue?

For the record, I'm also seeing that GraphFrames are broken on Python 3 as of Spark 2.0.0 / GraphFrames 0.2:

```
>>> print(spark.version)
2.0.0
>>> import sys
>>> print(sys.version)
3.5.2 (default, Sep 28 2016, 18:08:09) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)]
>>> from graphframes.examples import Graphs
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
zipimport.ZipImportError: can't find module 'graphframes'
```

This works fine for me on Python 2.7.12.

Also, if someone wants to tackle the issue of Python 3 compatibility, where would they start? I can't seem to find documentation on how to build GraphFrames for Python.

nchammas:
Looks like this has been addressed in a recent PR: #135

thunterdb:
@renewjoy this issue is fixed in master, and a new release will be done soon. I am going to close this ticket.

8lenka8:
Hi,
[...]

PR #135, opened 2016-11-10
python 3 support

thunterdb:


.travis.yml
```diff
@@ -2,31 +2,50 @@ language: scala
...
-    - SPARK_VERSION=1.6.2 SPARK_BUILD="spark-1.6.2-bin-hadoop2.6" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-1.6.2-bin-hadoop2.6.tgz"
+    - PYSPARK_PYTHON=python3 SPARK_VERSION=2.0.1 SPARK_BUILD="spark-2.0.1-bin-hadoop2.7" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-2.0.1-bin-hadoop2.7.tgz"
+    - PYSPARK_PYTHON=python3 SPARK_VERSION=1.6.2 SPARK_BUILD="spark-1.6.2-bin-hadoop2.6" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-1.6.2-bin-hadoop2.6.tgz"
+    - PYSPARK_PYTHON=python2 SPARK_VERSION=2.0.1 SPARK_BUILD="spark-2.0.1-bin-hadoop2.7" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-2.0.1-bin-hadoop2.7.tgz"
+    - PYSPARK_PYTHON=python2 SPARK_VERSION=1.6.2 SPARK_BUILD="spark-1.6.2-bin-hadoop2.6" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-1.6.2-bin-hadoop2.6.tgz"
+
+addons:
+  apt:
+    packages:
+     - python3
```
mengxr: It seems that Python3 should be preinstalled on trusty image: https://docs.travis-ci.com/user/trusty-ci-environment/#Python. Do we need it here?

.travis.yml
```diff
@@ -2,31 +2,50 @@ language: scala
...
+     - python3-pip
 
 matrix:
   exclude:
   - scala: 2.10.5 # Spark 2.0.0 pre-built with Scala 2.11 in the tgz download
-    env: SPARK_VERSION=2.0.1 SPARK_BUILD="spark-2.0.1-bin-hadoop2.7" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-2.0.1-bin-hadoop2.7.tgz"
```
mengxr: Is it simpler if we don't list `scala` as the main language and let the test line be `sbt -Dscala.version=2.10.6`, e.g.?

.travis.yml
```diff
@@ -2,31 +2,50 @@ language: scala
...
   - scala: 2.11.7
-    env: SPARK_VERSION=1.6.2 SPARK_BUILD="spark-1.6.2-bin-hadoop2.6" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-1.6.2-bin-hadoop2.6.tgz"
+    env: PYSPARK_PYTHON=python2 SPARK_VERSION=1.6.2 SPARK_BUILD="spark-1.6.2-bin-hadoop2.6" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-1.6.2-bin-hadoop2.6.tgz"
 
 before_install:
  - ./bin/download_travis_dependencies.sh
 
 install:
  - pip install --user -r ./python/requirements.txt
+ - pip3 install --user -r ./python/requirements.txt
```
mengxr: We might use `pyenv` to switch python version first and make this line the same for py2 and py3.

.travis.yml
mengxr: Btw, shall we also bump scala version from `2.10.5` to `2.10.6`?

.travis.yml
thunterdb: You are right, I have not tried to remove these flags since moving to trusty.

.travis.yml
thunterdb: I am going to try it out.

.travis.yml
```diff
@@ -1,34 +1,31 @@
-language: scala
-scala:
-  - 2.10.5
-  - 2.11.7
 jdk:
   - oraclejdk7 # openJDK crashes sometimes
-sudo: false
+
+sudo: required
```
mengxr: If we no longer need to install `pip`, does it mean we don't need `sudo` either?

thunterdb: We do. trusty is beta, and fails (jvm package not properly installed)

.travis.yml
thunterdb: I suggest we keep it this way for now. It just takes a few seconds and it is simpler to install both.

nchammas:
I suppose this addresses #85.

In the future, it would be good to include "Fixes #85" in the PR description so that when the PR gets merged in, the associated issue gets automatically closed and people are notified that their issue is fixed.

thunterdb:
That's a great suggestion. Yes, we should tag the ticket in the title.

PR #136, opened 2016-11-10
Set default checkpoint interval to 2 in connected components

mengxr:
I tested the implementation on a relatively large graph. It seems setting the checkpoint interval to 2 hit the balance between checkpoint overhead and complexity of query plan optimization. This is what I got:

* =1: 11.2 mins, 50 jobs
* =2: 8.6 mins, 42 jobs
* =3: ran out of memory after 30 mins ...

cc: @thunterdb

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -94,12 +94,12 @@ class ConnectedComponents private[graphframes] (
    */
   def getAlgorithm: String = algorithm
 
-  private var checkpointInterval: Int = 1
+  private var checkpointInterval: Int = 2
 
   /**
-   * Sets checkpoint interval in terms of number of iterations (default: 1).
-   * Checkpointing regularly helps recover from failures, clean shuffle files, and shorten the
-   * lineage of the computation graph.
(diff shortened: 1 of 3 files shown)


src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -94,12 +94,12 @@ class ConnectedComponents private[graphframes] (
...
-  private var checkpointInterval: Int = 1
+  private var checkpointInterval: Int = 2
 
   /**
-   * Sets checkpoint interval in terms of number of iterations (default: 1).
-   * Checkpointing regularly helps recover from failures, clean shuffle files, and shorten the
-   * lineage of the computation graph.
+   * Sets checkpoint interval in terms of number of iterations (default: 2).
+   * Checkpointing regularly helps recover from failures, clean shuffle files, shorten the
+   * lineage of the computation graph, and reduce the complexity of plan optimization.
```
thunterdb: Given the exponential blowup that we observed, can you also include a rule of thumb such as 'more than 5 will put considerable pressure on the SQL optimizer' or something like that?

mengxr: Updated the doc and inserted a warning.

thunterdb:
LGTM, feel free to merge.

PR #138, opened 2016-11-11
prepare release v0.3.0

mengxr:
Prepare release 0.3:
* update README, versions
* update API doc in CC
* hide protected methods in `Logging`

PR #151, opened 2016-12-21
Re-add note about Python 3 support

nchammas:
I talked @mengxr out of adding this in the first time around [here in this PR](https://github.com/graphframes/graphframes/pull/138), but it looks like GrapheFrames works fine with Python 3, and I've been using it as such for many weeks now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,4 +35,5 @@ We welcome open source contributions as well!
 - 0.2.0 release for Spark 2.0 (work of @felixcheung)
 - 0.3.0
   - DataFrame-based connected components implementation
+  - added support for Python 3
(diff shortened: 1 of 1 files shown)


mengxr:
@nchammas I think we cannot promise python 3 support without testing it in Travis CI. Do you have time to update the Travis file?

nchammas:
Good point. Will do.

nchammas:
@mengxr - It looks like tests are already being [run against Python 3](https://github.com/graphframes/graphframes/blob/775367b8a96941f21e0a76bb628771adb0771ff0/.travis.yml#L18-L19).

PR #153, opened 2016-12-30
[157] Support Spark 2.1.0

felixcheung:
add support for 2.1.0 while keeping 2.0.2


src/main/scala/org/graphframes/examples/Graphs.scala
```diff
@@ -184,8 +184,11 @@ class Graphs private[graphframes] () {
     require(n >= 1, s"Grid graph must have size >= 1, but was given invalid value n = $n")
     val rows = sqlContext.createDataFrame(Range(0, n).map(i => Tuple1(i))).toDF("i")
     val cols = rows.select(rows("i").as("j"))
+
     // Cartesian join to create grid
-    val coordinates = rows.join(cols)
+    // Handle SPARK-15425: check Spark version for >1.6.3 (ie. 2.x) to call crossJoin explicitly
```
felixcheung: I added the comment with text `1.6.3` so that it would be easy to find

thunterdb:
It looks good. If you look at the output of the test script, though, the code gen fails in one case. I will  open a separate ticket to investigate the issue.

PR #154, opened 2016-12-30
Script to read up-to-date version string for python api documentation

felixcheung:
See header at http://graphframes.github.io/api/python/index.html
It should match the package version which is 0.3.0, or 0.4.0-SNAPSHOT

Changes:
update python doc version from config instead
fix to work with new py4j in Spark 2.1
add switch to skip python doc


thunterdb:
@felixcheung thanks for chasing these dependencies. I will push the website after it merges.

PR #161, opened 2017-01-06
[160] ConnectedComponents checkpointing should construct FileSystem using Path

### LINKED ISSUE #160: ConnectedComponents checkpointing should construct FileSystem using Path
See https://www.mail-archive.com/user@spark.apache.org/msg60669.html for the original bug report.  I recommend we modify FileSystem constructors to take Path + hadoop config, rather than only the hadoop config.  That should be more robust AFAIK.

felixcheung:
to make sure we are using the right FileSystem.
took clue from https://github.com/apache/spark/blob/master/streaming/src/main/scala/org/apache/spark/streaming/dstream/DStreamCheckpointData.scala#L88

fix #160

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -156,8 +156,6 @@ object ConnectedComponents extends Logging {
   private val ALGO_GRAPHX = "graphx"
   private val ALGO_GRAPHFRAMES = "graphframes"
 
-  @transient private var fileSystem: FileSystem = null
-
   /**
    * Supported algorithms in [[org.graphframes.lib.ConnectedComponents.setAlgorithm]]: "graphframes"
    * and "graphx".
@@ -339,10 +337,7 @@ object ConnectedComponents extends Logging {
         // remove previous checkpoint
         if (iteration > checkpointInterval) {
           val path = new Path(s"${checkpointDir.get}/${iteration - checkpointInterval}")
-          if (fileSystem == null) {
-            fileSystem = path.getFileSystem(sc.hadoopConfiguration)
-          }
-          fileSystem.delete(path, true)
+          path.getFileSystem(sc.hadoopConfiguration).delete(path, true)
         }
 
         System.gc() // hint Spark to clean shuffle directories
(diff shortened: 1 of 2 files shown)


src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -336,8 +338,11 @@ object ConnectedComponents extends Logging {
 
         // remove previous checkpoint
         if (iteration > checkpointInterval) {
-          FileSystem.get(sc.hadoopConfiguration)
-            .delete(new Path(s"${checkpointDir.get}/${iteration - checkpointInterval}"), true)
+          val path = new Path(s"${checkpointDir.get}/${iteration - checkpointInterval}")
+          if (fileSystem == null) {
```
thunterdb: how expensive is it to obtain a file system object? I suspect it is negligible compared to the cost of other Spark operations. Let's us not store it unless it is necessary.

thunterdb:
@felixcheung thanks, I have a comment

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
felixcheung: From what I can see
- DStreamCheckpointData.scala does it this way but its history is hard to track down since it is moved in 2014
- this is what the call [leads to](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3285) - basically it hits a lock multiple times mostly to cached the returned FileSystem

Since it is caching already let's move to not store it.

felixcheung: although to fix #162, we will likely need to store the FileSystem in order to clean up the very last checkpoint saved.

but more can come then

thunterdb:
One of the tests ran out of heap space. We should be able to fix that in `python/run-tests.sh` with something like that:
http://stackoverflow.com/questions/31464727/increase-memory-available-to-pyspark-at-runtime

felixcheung:
all passed

jkbradley:
LGTM, thank you!

PR #163, opened 2017-01-09
Remove backported pagerank algorithm

aray:
Since GraphFrames 0.3.0 removed support for Spark 1.4 and 1.5, this backport is no longer needed.


felixcheung:
that's great - could you check if it works with Spark 2.1. I tried looking into this earlier and found a lot of non-trivial changes while working on #153

aray:
I merged your 2.1 branch and verified all unit tests pass. Per `git diff v2.0.0..v2.1.0 -- /Users/andrew/spark/graphx/src/main/scala/org/apache/spark/graphx/lib/PageRank.scala` the only non trivial changes were in the addition of a new parallel personalized pagerank algorithm that is a completely separate method than what we use.

thunterdb:
Thank you @aray. I restarted the build to check against the latest master.

thunterdb:
Merging.

Issue #57, opened 2016-03-10
Add aggregateMessages API to Python


jkbradley:
It should mimic the Scala API and will likely need to be re-implemented in Python, following the Scala implementation.

hughchristensen:
@thunterdb @jkbradley @qingpeng @mengxr @felixcheung @tsivula

felixcheung:
I could help. Where are you at?

tsivula:
Hello, I'm in Finland at GMT+2. I'll dive into this on Monday. It would be nice if I could consult you through some channel. Thank you in advance for any help.

tsivula:
Hello @felixcheung, I just pushed my implementation proposition in [this commit](https://github.com/tsivula/graphframes/commit/dc9649fff1814a770f427556019fd11ce2863813). It would be nice if you could take a look at it.

I successfully tested it locally with a simple example:
```python
from graphframe import AggregateMessages as AM
from graphframes.examples import Graphs

g = Graphs(sqlContext).friends()  # Get example graph

df = g.aggregateMessages(
    pyspark.sql.functions.sum(AM.msg()),
    msgToDst=pyspark.sql.functions.lit(1))
```

rogerthealien:
Hi,

I'm very interested in this as well. @tsivula I'm trying to test your implementation. Is it possible to use a UDF as an aggregation function?

felixcheung:
@tsivula sorry about delay, I'm getting occupied this week and next.

I checked your commit, looking good! probably next would be to add some tests for it?

tsivula:
Hello, I just created PR #169 for this.

hughchristensen:
@thunterdb @jkbradley @qingpeng @mengxr @felixcheung Hi Guys. Can you help get this PR through please. We can not continue developing for this library until the PR has been accepted. 

Please can you do a best effort to get this done as we have dev resources waiting to go now. thank you!

felixcheung:
As signed off previously, this looks good to me.

Sorry someone else would need to jump in to merge this.

Issue #49, opened 2016-03-01
Consistency checks upon construction


jkbradley:
When GraphFrames are constructed, we do not check that the vertices DataFrame contains all vertices from the edges DataFrame.  We should do that somehow (ideally lazily).

This will likely be a problem for subgraph selection.

PR #169, opened 2017-02-05
[57] [170] aggregateMessages for Python

tsivula:
This PR adds `aggregateMessages` API to Python (issue #57)


felixcheung:
could you update the PR title to prefix `[57]` - this indicates this PR is fixing the issue # 57

python/graphframes/graphframe.py
```diff
@@ -17,6 +17,8 @@
 
 from pyspark import SparkContext
 from pyspark.sql import DataFrame, SQLContext
+from pyspark.sql.functions import col
+from pyspark.sql.column import Column
```
felixcheung: we would probably prefer to sort this by the namespace

python/graphframes/graphframe.py
```diff
@@ -205,6 +207,37 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+    def aggregateMessages(self, aggCol, msgToSrc=None, msgToDst=None):
+        """
+        Aggregates messages from the neighbours.
+
+        When specifying the messages and aggregation function, the user may reference columns using
+        the static methods in :class:`AggregateMessages`.
+
+        See Scala documentation for more details.
+
+        :param aggCol: `pyspark.sql.column.Column` for the requested aggregation output
```
felixcheung: could this be supported as SQL expression too?

python/graphframes/graphframe.py
```diff
@@ -205,6 +207,37 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+        if msgToSrc is not None:
+            if isinstance(msgToSrc, Column):
+                builder.sendToSrc(msgToSrc._jc)
+            else:
+                builder.sendToSrc(msgToSrc)
+        if msgToDst is not None:
+            if isinstance(msgToDst, Column):
+                builder.sendToDst(msgToDst._jc)
+            else:
+                builder.sendToDst(msgToDst)
```
felixcheung: should we require that either one of msgToSr or msgToDst is not None?

python/graphframes/graphframe.py
```diff
@@ -323,6 +356,50 @@ def triangleCount(self):
...
+    @staticmethod
+    def msg():
+        """Reference for message column, used for specifying aggregation function."""
+        jvm_gf_api = _java_api(SparkContext)
+        return col(jvm_gf_api.aggregateMessages().MSG_COL_NAME())
+
+    @staticmethod
+    def getCachedDataFrame(df):
+        """
+        Cache a DataFrame, and return the cached version.
```
felixcheung: should we emphasize that the returned one is a new DataFrame?

python/graphframes/graphframe.py
tsivula: Added check for this

python/graphframes/graphframe.py
tsivula: added support for String SQL expression in `AggregateMessages.agg()` in Scala and Python

src/main/scala/org/graphframes/lib/AggregateMessages.scala
```diff
@@ -130,6 +130,13 @@ class AggregateMessages private[graphframes] (private val g: GraphFrame)
     }
     unionedMsgs.groupBy(ID).agg(aggCol)
   }
+
+  /**
+   * Run the aggregation, specifying SQL expression as a String
+   *
+   * See the overloaded method documentation for more details.
+   */
+  def agg(aggCol: String): DataFrame = agg(expr(aggCol))
```
felixcheung: Would be great to add a Scala test for this - this helps mitigate the Codecov decrease too

felixcheung:
this is looking great, @thunterdb @jkbradley could you review this too?

tsivula:
I now updated the PR:
- added gridIsingModel example graph to Python
- added BeliefPropagation example (addresses #170 )
- updated user-guide
- updated scala test for AggregateMessages

python/graphframes/examples.py
```diff
@@ -57,3 +60,216 @@ def friends(self):
         ], ["src", "dst", "relationship"])
         # Create a GraphFrame
         return GraphFrame(v, e)
+
+    def gridIsingModel(self, n, vStd=1.0, eStd=1.0):
+        """Grid Ising model with random parameters.
```
felixcheung: this is quite a bit we are adding here, would it be better if this is a separate file like BeliefPropagation.scala is a separate file?

python/graphframes/graphframe.py
```diff
@@ -16,7 +16,7 @@
 #
 
 from pyspark import SparkContext
-from pyspark.sql import Column, DataFrame, functions as sqlfunctions, SQLContext
+from pyspark.sql import Column, DataFrame, functions as sqlFuncs, SQLContext
```
felixcheung: I actually think `sqlfunctions` make more sense? is there a reason for this change?

python/graphframes/examples.py
tsivula: I also think this module should be split into subpackage `examples` with modules `Graphs` and `BeliefPropagation`, just as in Scala side. I just did not have courage to do it without clarification from you. I'll do it.

python/graphframes/graphframe.py
tsivula: No reason (maybe trying to make it shorter). I see it is not good. I'll change it back. Or would you prefer some of these?
- `from pyspark import sql` and just call e.g. `sql.functions.col` and  `sql.functions.sum`
- `from pyspark.sql.functions import col, sum as sqlSum` or `sql_sum`

felixcheung: I think having sqlfunctions and having calls as `sqlfunctions.col` `sqlfunctions.sum` make sense

jkbradley:
Thanks a lot @tsivula for the PR and @felixcheung for reviewing!  I'll try to take a look soon, and I'm also going to ask @mrbago to check it out.

MrBago:
@tsivula Thanks for the PR, this look really good. I have a few thoughts I'll drop in line.

python/graphframes/graphframe.py
MrBago: Can you update the docstring to reflect the added SQL expression support.

python/graphframes/tests.py
```diff
@@ -204,3 +237,67 @@ def test_triangle_counts(self):
...
+        results = BeliefPropagation.runBPwithGraphFrames(g, numIter)
+
+        # check beliefs are valid
+        for row in results.vertices.select('belief').collect():
+            belief = row['belief']
+            self.assertTrue(
+                0 <= belief <= 1,
+                msg="Expected belief to be probability in [0,1], but found {}".format(belief))
+
+    def test_graph_friends(self):
```
MrBago: This test doesn't feel very informative. The "fiends" are hardcoded in both the example and the test so this test is mostly testing if someone has changed the package source.

docs/user-guide.md
```diff
@@ -859,14 +859,40 @@ val agg = g.aggregateMessages
...
+    sqlsum(AM.msg()).alias("summedAges"),
+    msgToSrc=msgToSrc,
+    msgToDst=msgToDst)
+agg.show()
+
+{% endhighlight %}
 
 For a more complex example, look at the code used to implement the
-[Belief Propagation example](api/scala/index.html#org.graphframes.examples.BeliefPropagation$).
+[Belief Propagation example](api/python/graphframes.examples.html#graphframes.examples.BeliefPropagation).
```
felixcheung: I think you might have ended up taken this out for the Scala tab - can you check?

tsivula: Checked, It's there on line 864. The diff is just a bit confusing.

python/graphframes/examples.py
```diff
@@ -57,3 +60,216 @@ def friends(self):
...
+
+        # run BP for numIter iterations
+        for iter_ in range(numIter):
+            # for each color, have that color receive messages from neighbors
+            for color in range(numColors):
+                # Send messages to vertices of the current color.
+                # We may send to source or destination since edges are treated as undirected.
+                msgForSrc = sqlFuncs.when(
+                    AM.src()['color'] == color,
+                    AM.edge()['b'] * AM.dst()['belief'])
```
jkbradley: +1 for consistency with Scala

python/graphframes/examples.py
```diff
@@ -15,14 +15,17 @@
 # limitations under the License.
 #
 
-from .graphframe import GraphFrame
+import itertools
+import math
+from pyspark import SparkConf, SparkContext
```
jkbradley: Separate imports (Python, Spark, graphframes), just as we'd do in Spark itself

src/test/scala/org/graphframes/lib/AggregateMessagesSuite.scala
```diff
@@ -56,8 +57,24 @@ class AggregateMessagesSuite extends SparkFunSuite with GraphFrameTestSparkConte
       }
       a.toMap
     }
+    // Compare to the true values.
     aggMap.keys.foreach { case user =>
       assert(aggMap(user) === trueAgg(user), s"Failure on user $user")
     }
+    // Perfom the aggregation againg, this timne providing the messages as Strings instead.
```
jkbradley: typos:
 againg -> again
 timne -> time

python/graphframes/examples.py
tsivula: I thought about this also.

In Scala we can call function `c = AM.src` without parenthesis and get a `Column` instance. Furthermore, the returned `Column` instance is indexed with `c2 = c('color')`.

In Python functions are not (typically) called without parenthesis so we need to do `c = AM.src()`. Also the indexing is done with `c2 = c['color']`.

The empty parenthesis in Python can be dropped out with a hack but this is not conventional. Class `AM` is just for grouping purposes to make the structure consistent with Scala and it is not initialised. If `AM` would be initialised, `@property` decorator could be used to allow calling `AM.src` without parenthesis. I can make some custom modifications to allow applying similar decorator to create a "class property". However, I believe the concept of properties in python is meant for different purpose and it would become confusing. I would not recommend it.

What do you think? Did I get something wrong?

tsivula:
Thanks for the feedback, I addressed the comments and refactored the `examples.py`. Considering this one comment https://github.com/graphframes/graphframes/pull/169#discussion-diff-108307377R233, I did not make any changes yet.

python/graphframes/examples.py
MrBago: I think you're right about the indexing, we shouldn't mess with that, but do we need `src` and `dst` to be methods? Can they simply be attributes of the class? Sorry if I missed something obvious, but can we do something like:
```
_jvm_gf_api = _java_api(SparkContext)
class AggregateMessages(object):
    # Reference for source column, used for specifying messages."
    src = sqlfunctions.col(_jvm_gf_api.SRC())
    # Reference for destination column, used for specifying messages."
    dst = sqlfunctions.col(_jvm_gf_api.DST())
```

If that doesn't work, I don't think there is anything hackie about initializing the AggregateMessages as a singleton object and using properties.

tsivula: In scala, I believe, these return new Column instances in every access. It is better to do the same in Python also. This avoids accidentally messing up with it by modifying the obtained Column object in-place.

BTW. I found out that building the python API doc gets complicated if there are such class or module level attributes that rely on the graphframes java api. The current build setup fails here as the import fails. I found this when I tried to make `GraphFrame.ID` etc. into class attributes.

Bellow are some candidates; whould some of these be good?

### 1)
Called by `AM().src`
```python
class AggregateMessages(object):
    def __init__(self):
        self._src_col_name = _java_api(SparkContext).SRC()
    @property
    def src(self):
        return sqlfunctions.col(self._src_col_name)
```
Simple alternative to the current `AM.src()`

### 2)
Called by `AM.src`
```python
class _ClassProperty(property):
    def __get__(self, cls, owner):
        return self.fget.__get__(None, owner)()

class AggregateMessages(object):
    @_ClassProperty
    @staticmethod
    def src(self):
        return sqlfunctions.col(_java_api(SparkContext).SRC())
```
No parenthesis needed, this is the one I spoke of earlier

### 3)
Called by `AM.src`
```python
class _AggregateMessages(object):
    @property
    def src(self):
        return sqlfunctions.col(_java_api(SparkContext).SRC())

AggregateMessages = _AggregateMessages()
```
No parenthesis needed, maybe controversial http://stackoverflow.com/a/6760726

### 4)
Called by `AM.src`
```python
def singleton_object(cls):
    """Class decorator that transforms (and replaces) a class definition (which
    must have a Singleton metaclass) with the actual singleton object. Ensures
    that the resulting object can still be "instantiated" (i.e., called),
    returning the same object. Also ensures the object can be pickled, is
    hashable, and has the correct string representation (the name of the
    singleton)
    """
    assert isinstance(cls, _Singleton), \
        cls.__name__ + " must use Singleton metaclass"

    def self_instantiate(self):
        return self

    cls.__call__ = self_instantiate
    cls.__hash__ = lambda self: hash(cls)
    cls.__repr__ = lambda self: cls.__name__
    cls.__reduce__ = lambda self: cls.__name__
    obj = cls()
    obj.__name__ = cls.__name__
    return obj

class _Singleton(type):
    """ A metaclass that creates a Singleton base class when called. """
    _instances = {}
    def __call__(cls, *args, **kwargs):
        if cls not in cls._instances:
            cls._instances[cls] = super(_Singleton, cls).__call__(*args, **kwargs)
        return cls._instances[cls]

class Singleton(_Singleton('SingletonMeta', (object,), {})): pass

@singleton_object
class AggregateMessages(Singleton):
    @property
    def src(self):
        return sqlfunctions.col(_java_api(SparkContext).SRC())

```
I managed to do this singleton implementation by using parts from: https://michaelgoerz.net/notes/singleton-objects-in-python.html and http://stackoverflow.com/questions/6760685/creating-a-singleton-in-python
I tried that it should work for both python 2 and 3.

MrBago: I vote for 2 or 3, both look good to me.

Maybe I took the conversation off-track by using the word "singleton", sorry. We don't have the issues normally associated with singletons in the PR because there is no reason for us to try and prevent the user from initializing their own version of `_AggregateMessages`. If some user wants to do that for some reason, it's perfectly fine. However, I think we should provide an initialized version for most users who don't need to know the implementation details, they simply need a name space that'll allow them to use `AM.src` and `AM.dst`. I think both option 2 and option 3 do that quite well.

tsivula:
I made the one change discussed in the comments here https://github.com/graphframes/graphframes/pull/169#discussion-diff-108307377R233

python/graphframes/examples.py
felixcheung: thanks for the very detailed analysis.
from my understanding this only returns a reference, not an actual column that should update each time? `sqlfunctions.col(_java_api(SparkContext).SRC())` - so in that case is it not possible to pre-initialize it?

in any case, I quite like 2 - is there a reason we need a custom decorator and not just `@property` should a bit more code?

felixcheung:
thanks! have a question here https://github.com/graphframes/graphframes/pull/169#discussion_r111826038

docs/user-guide.md
felixcheung: right!

python/graphframes/examples.py
MrBago: When you normally create a property, accessing the property through the instance causes the property to evaluate, while accessing it through the class returns the property object proper. For example:

```
In [1]: class Foo(object):
   ...:     @property
   ...:     def foo(self):
   ...:         return "result of property evaluation"
   ...:

In [2]: Foo().foo
Out[2]: 'result of property evaluation'

In [3]: Foo.foo
Out[3]: <property at 0x1023d4520>
```

Here we want to use a "static method property" or `_ClassProperty`. @tsivula has found a nice way to modify the property class to get the behavior we want.

tsivula: The corresponding Scala method returns new instance with every call. I think this is good and safe so that modifying the returned instace in place (maybe accidentally) does not affect future calls.

felixcheung: thanks! make sense

felixcheung:
@MrBago would you like to review another pass?

python/graphframes/examples/belief_propagation.py
```diff
@@ -0,0 +1,173 @@
...
+    def runBPwithGraphFrames(cls, g, numIter):
+        """Run Belief Propagation using GraphFrame.
+
+        This implementation of BP shows how to use GraphFrame's aggregateMessages method.
+        """
+        # choose colors for vertices for BP scheduling
+        colorG = cls._colorGraph(g)
+        numColors = colorG.vertices.select('color').distinct().count()
+
+        # TODO: handle vertices without any edges
```
felixcheung: hmm, should this be here?

felixcheung:
opened #187 - think it'll be good to have a way to run examples (scala/python) in Travis CI runs

python/graphframes/examples/belief_propagation.py
```diff
@@ -0,0 +1,173 @@
...
+    g.edges.show()
+
+    # run BP for 5 iterations
+    numIter = 5
+    results = BeliefPropagation.runBPwithGraphFrames(g, numIter)
+
+    # display beliefs
+    beliefs = results.vertices.select('id', 'belief')
+    print("Done with BP. Final beliefs after {} iterations:".format(numIter))
+    beliefs.show()
```
felixcheung: does this result match Scala's?

python/graphframes/examples/belief_propagation.py
tsivula: This comment is also in the scala side.
In our example, using the gridIsingModel example graph, the graph should not contain such vertices anyway.

python/graphframes/examples/belief_propagation.py
tsivula: For some reason, the `pyspark.sql.functions.randn` gives different random numbers compared to the scala version, evethough the seed is the same. Thus the original example graph is a bit different and the outcome is also different. Should this be inspected somehow?

P.S. oddly from the 9 random draws, first two are identical but the rest are not

tsivula: I get this for python:
```
Original Ising model:
+---+---+---+-------------------+
|  i|  j| id|                  a|
+---+---+---+-------------------+
|  0|  0|0,0| -1.694613256968788|
|  0|  1|0,1| 0.8810493517315586|
|  0|  2|0,2| 0.9050342451284052|
|  1|  0|1,0| 0.6750384288359632|
|  1|  1|1,1| 2.2217120063656193|
|  1|  2|1,2|-0.3807804767931157|
|  2|  0|2,0|-1.0768226745614016|
|  2|  1|2,1| 0.6439379854718037|
|  2|  2|2,2| -2.517736032719354|
+---+---+---+-------------------+

+---+---+-------------------+
|src|dst|                  b|
+---+---+-------------------+
|0,0|0,1| 0.9050342451284052|
|0,1|0,2| 0.6750384288359632|
|1,0|1,1| 2.2217120063656193|
|1,1|1,2|-1.0768226745614016|
|2,0|2,1|0.35452949087488994|
|2,1|2,2|  0.867351146019913|
|0,0|1,0|0.22614416575710014|
|0,1|1,1| 0.2143789015435314|
|0,2|1,2|-1.1718358468823387|
|1,0|2,0| 1.3133477706235883|
|1,1|2,1|-2.3338927203206326|
|1,2|2,2|-1.0878763030720948|
+---+---+-------------------+
```
and this for Scala
```
Original Ising model:
+---+---+---+-------------------+
|  i|  j| id|                  a|
+---+---+---+-------------------+
|  0|  0|0,0| -1.694613256968788|
|  0|  1|0,1| 0.8810493517315586|
|  0|  2|0,2|-0.5166048414445207|
|  1|  0|1,0|-1.1000837184563204|
|  1|  1|1,1|-0.5391809043782484|
|  1|  2|1,2| 1.0386363349029881|
|  2|  0|2,0|0.12177090579382759|
|  2|  1|2,1|-0.8288946004978055|
|  2|  2|2,2| 1.1492010250003564|
+---+---+---+-------------------+

+---+---+-------------------+
|src|dst|                  b|
+---+---+-------------------+
|0,0|0,1| 0.9050342451284052|
|0,1|0,2| 0.6750384288359632|
|1,0|1,1| 2.2217120063656193|
|1,1|1,2|-1.0768226745614016|
|2,0|2,1|0.35452949087488994|
|2,1|2,2|  0.867351146019913|
|0,0|1,0|0.22614416575710014|
|0,1|1,1| 0.2143789015435314|
|0,2|1,2|-1.1718358468823387|
|1,0|2,0| 1.3133477706235883|
|1,1|2,1|-2.3338927203206326|
|1,2|2,2|-1.0878763030720948|
+---+---+-------------------+
```

python/graphframes/examples/belief_propagation.py
felixcheung: ok, think this is #49

python/graphframes/examples/belief_propagation.py
felixcheung: hmm, `pyspark.sql.functions.randn` is different might be worthwhile to look into in Spark

felixcheung:
hmm, CI failure seems weird but related? could you take a look please?

python/graphframes/examples/belief_propagation.py
tsivula: Hmm, I think this issue #49 is not related to this. #49 considers if the graph definition is broken. I believe this todo considers if the function is used with a valid graph that has some vertices with degree 0. In our example such graph is not provided for it but user might try it with such a graph.

I'll see if I could add simple check for those and deal with them.

tsivula:
Ah sorry. I only run the tests with python3. In python2 the last commit resulted in a circular import error.

Explanation:
Because module `graphframes/examples/belief_propagation.py` should be submittable to `spark-submit` it needs to import `graphframes/examples` which imports `graphframes/examples/belief_propagation.py` in `graphframes/examples/__init__.py`. I had `from graphframes import examples`, which is fine in my version of Python 3. I changed this to `import graphframes.examples`, which seems to be ok also for Python 2.7.12.

python/graphframes/examples/belief_propagation.py
tsivula: I tested the BeliefPropagation function by submitting similar identical manually created graph to both Python and Spark implementations and got identical results.

Also I noticed this comment in the Scala API docs of `org.apache.spark.sql.functions.randn`
```
Note    This is indeterministic when data partitions are not fixed.
```
I suppose this is the reason for getting different random numbers.

felixcheung: ah ok, thanks for checking

python/graphframes/examples/belief_propagation.py
felixcheung: I see, that would be specific to the logic with BP then. We could follow up on that

felixcheung:
merged, thanks!

tsivula:
Thanks for the review and comments. May I enquire what kind of plans you have for the schedule of the next release?

felixcheung:
as discussed in PR 189, we are going to roll a release with the fix?

python/graphframes/graphframe.py
```diff
@@ -205,6 +205,44 @@ def bfs(self, fromExpr, toExpr, edgeFilter=None, maxPathLength=10):
...
+        :return: DataFrame with columns for the vertex ID and the resulting aggregated message
+        """
+        # Check that either msgToSrc, msgToDst, or both are provided
+        if msgToSrc is None and msgToDst is None:
+            raise ValueError("Either `msgToSrc`, `msgToDst`, or both have to be provided")
+        builder = self._jvm_graph.aggregateMessages()
+        if msgToSrc is not None:
+            if isinstance(msgToSrc, Column):
+                builder.sendToSrc(msgToSrc._jc)
+            else:
```
jkbradley: We should check the type to make sure it is String here.

python/graphframes/graphframe.py
```diff
@@ -323,6 +361,64 @@ def triangleCount(self):
...
+        """Reference for message column, used for specifying aggregation function."""
+        jvm_gf_api = _java_api(SparkContext)
+        return sqlfunctions.col(jvm_gf_api.aggregateMessages().MSG_COL_NAME())
+
+    @staticmethod
+    def getCachedDataFrame(df):
+        """
+        Create a new cached copy of a DataFrame.
+
+        This utility method is usefull for iterative DataFrame-based algorithms. See Scala
```
jkbradley: typo: "useful"

jkbradley:
Yeah, I'd like to do a release soon for that bug, and it can include this as well.  When we do the release, I'll initiate creating release branches so that we can have actual patch releases in the future (instead of just doing new minor releases).

Maybe I'll wait to drop 1.6 support until after this release.

tsivula:
I can do these during the weekend.

python/graphframes/graphframe.py
tsivula: About this, we had this conversation with @felixcheung earlier here https://github.com/graphframes/graphframes/pull/169#commitcomment-20708603

Do you have any advices for this?

jkbradley: Oh I missed that earlier.  In Spark, we generally do this: https://github.com/apache/spark/blob/e3d2022e4b73c5bc352eb001ed334d91ce36109c/python/pyspark/ml/feature.py#L18

and: https://github.com/apache/spark/blob/e3d2022e4b73c5bc352eb001ed334d91ce36109c/python/pyspark/ml/feature.py#L2833

Could you do that approach please?

felixcheung:
I noticed also in other places like https://github.com/graphframes/graphframes/commit/490f51a2995d0ddfd55797cc442104097f514e43#diff-e81e6b169c0aa35012a3263b2f31b330R70
we have `assert` instead of `raise Error` - I think it'll be good to make it consistent

tsivula:
I now implemented the discussed changes in PR #196

PR #173, opened 2017-02-13
Python packaging

Orhideous:
Allow to wrap code into ordinary python package so it can be installed with `pip`.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,11 +1,11 @@
 # Test
-nose>=1.3
+nose==1
 
 # Lint
-pep8>=1.7
-flake8>=2.5
-pyflakes>=1.2
+pep8==1
+flake8==2
+pyflakes==1
 
 # Packaging
-wheel>=0.29
-setuptools>=26.1
+wheel==0.29
+setuptools==26
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,2 +1,11 @@
-# This file should list any python package dependencies.
-nose==1.3.3
+# Test
+nose>=1.3
+
+# Lint
+pep8>=1.7
+flake8>=2.5
+pyflakes>=1.2
+
+# Packaging
+wheel>=0.29
+setuptools>=26.1
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,2 +1,27 @@
-# Your python setup file. An example can be found at:
-# https://github.com/pypa/sampleproject/blob/master/setup.py
+#!/usr/bin/env python
+# coding=utf-8
+
+from setuptools import setup
+from setuptools import find_packages
+
+
+setup(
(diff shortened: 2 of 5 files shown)


python/requirements.txt
```diff
@@ -1,2 +1,11 @@
...
\ No newline at end of file
+# Test
+nose>=1.3
+
+# Lint
+pep8>=1.7
+flake8>=2.5
+pyflakes>=1.2
+
+# Packaging
```
thunterdb: I am not very familiar with the python ecosystem, but in java, bounds like this can be a cause for issue due to API breakages. Can you add some upper bounds as well?

python/setup.py
```diff
@@ -1,2 +1,27 @@
-# Your python setup file. An example can be found at:
-# https://github.com/pypa/sampleproject/blob/master/setup.py
+#!/usr/bin/env python
+# coding=utf-8
+
+from setuptools import setup
+from setuptools import find_packages
+
+
+setup(
```
thunterdb: Thanks!

python/requirements.txt
Orhideous: @thunterdb This packages are pretty stable by now… but in case of API changes I pinned theirs major versions.

python/setup.py
```diff
@@ -1,2 +1,27 @@
...
+    name="graphframes",
+    version="0.3.0",
```
felixcheung: can you update this to 0.5.0  - btw, is there a way to get this programmatically so we don't forget to update this for each release?

python/setup.py
```diff
@@ -1,2 +1,27 @@
...
+    license="ASF 2.0",
+    zip_safe=False,
+    keywords=["graphframes", "spark", "graph", "dataframe"],
+    classifiers=[
+        "Development Status :: 3 - Alpha",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Natural Language :: English",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.4",
```
felixcheung: add python 2?

python/requirements.txt
```diff
@@ -1,2 +1,11 @@
-# This file should list any python package dependencies.
-nose==1.3.3
\ No newline at end of file
+# Test
+nose==1
```
felixcheung: is there any issue moving from nose 1.3.3 to 1?

felixcheung: I think the main concern is with 1.2.x- not being really compatible with 1.3.3+

felixcheung:
@Orhideous ?

dataframing:
@felixcheung Is this still an open issue for GraphFrames? If so, I'm open to try addressing it.

PR #180, opened 2017-03-28
[157][166] Updated docs for 0.4.0 doc preparation

jkbradley:
* Removed "SNAPSHOT" label in config.yml for docs
* Updated ruby script for building Python docs to fail more loudly if SPARK_HOME is not present (which I tested manually) and to clean the build (which prevents build failures when built docs already exist)
* Added small clarification to user guide

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -14,7 +14,7 @@ include:
 
 # These allow the documentation to be updated with newer releases
 # of Spark, Scala, and Mesos.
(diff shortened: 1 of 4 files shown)


jkbradley:
This PR is paired with https://github.com/graphframes/graphframes.github.io/pull/7 which updates the actual github.io documentation.

jkbradley:
I'll merge this now since it seems to have worked.  Thanks @felixcheung !

Issue #76, opened 2016-04-04
Check algorithm parameters for validity


jkbradley:
For the stdlib algorithms (PageRank, etc.), we should check for valid parameter values in the setters, rather than failing somewhere internally.

bcajes:
Hi, in the process of familiarizing myself with this library, I've decided give this one a shot.  Let me know if the following is more or less what is needed:  
https://github.com/graphframes/graphframes/compare/master...bcajes:master 
If so, I'll add checks for the remaining algorithms.

jkbradley:
Yep, adding checks in the setter methods is right.  Just one comment about tests: Can you please use this pattern which we use in Spark itself?
```
withClue("explanation...") {
  intercept[MyExceptionType] {
    // code which causes exception
  }
}
```
Thanks!

PR #182, opened 2017-04-10
Checking algorithm parameters for validity

bcajes:
Adding some checks for parameters of various algorithms within GraphFrames #76 .  TODO: check to see if SVD++ parameter requirements can be made more specific.

### DIFF
--- a/src/main/scala/org/graphframes/lib/PageRank.scala
+++ b/src/main/scala/org/graphframes/lib/PageRank.scala
@@ -80,16 +80,20 @@ class PageRank private[graphframes] (
 
   /** Reset probability "alpha" */
   def resetProbability(value: Double): this.type = {
+    require(value >= 0 && value <= 1, s"Random reset probability must belong" +
+      s" to [0, 1], but got ${value}")
     resetProb = Some(value)
     this
   }
 
   def tol(value: Double): this.type = {
+    require(value >= 0, s"Tolerance must be no less than 0, but got ${value}")
     tol = Some(value)
     this
   }
 
   def maxIter(value: Int): this.type = {
+    require(value > 0, "maxIter should be > 0")
     maxIter = Some(value)
     this
   }
--- a/src/main/scala/org/graphframes/lib/ShortestPaths.scala
+++ b/src/main/scala/org/graphframes/lib/ShortestPaths.scala
@@ -45,7 +45,7 @@ class ShortestPaths private[graphframes] (private val graph: GraphFrame) extends
    * The list of landmark vertex ids. Shortest paths will be computed to each landmark.
    */
   def landmarks(value: Seq[Any]): this.type = {
-    // TODO(tjh) do some initial checks here, without running queries.
+    require(value.nonEmpty, "expecting at least one landmark")
     lmarks = Some(value)
     this
   }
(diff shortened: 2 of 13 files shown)


src/main/scala/org/graphframes/lib/ShortestPaths.scala
```diff
@@ -45,7 +45,7 @@ class ShortestPaths private[graphframes] (private val graph: GraphFrame) extends
    * The list of landmark vertex ids. Shortest paths will be computed to each landmark.
    */
   def landmarks(value: Seq[Any]): this.type = {
-    // TODO(tjh) do some initial checks here, without running queries.
+    require(value.nonEmpty, "expecting at least one landmark")
```
felixcheung: also check for the type within the Seq?

src/main/scala/org/graphframes/lib/PageRank.scala
```diff
@@ -80,16 +80,20 @@ class PageRank private[graphframes] (
...
   /** Reset probability "alpha" */
   def resetProbability(value: Double): this.type = {
+    require(value >= 0 && value <= 1, s"Random reset probability must belong" +
+      s" to [0, 1], but got ${value}")
     resetProb = Some(value)
     this
   }
 
   def tol(value: Double): this.type = {
+    require(value >= 0, s"Tolerance must be no less than 0, but got ${value}")
```
felixcheung: I'd leave this as `tol` for consistency with other checks?

src/main/scala/org/graphframes/lib/PageRank.scala
```diff
@@ -80,16 +80,20 @@ class PageRank private[graphframes] (
 
   /** Reset probability "alpha" */
   def resetProbability(value: Double): this.type = {
+    require(value >= 0 && value <= 1, s"Random reset probability must belong" +
```
felixcheung: I'd leave this as `resetProbability`?

src/main/scala/org/graphframes/lib/ShortestPaths.scala
bcajes: Tracing through this implementation, and correct me if I'm wrong, it looks like graphframe vertex IDs can be of any type allowable within a spark Dataframe.  I can add a pattern match check for all of those types.

src/main/scala/org/graphframes/lib/PageRank.scala
bcajes: okay

src/main/scala/org/graphframes/lib/PageRank.scala
bcajes: okay

bcajes:
@felixcheung Agreed some documentation and default values would be helpful.  I can add most of them.  Though, I'm not familiar enough with the gamma parameters of SVDPlusPlus.  So, if you or someone else can shed light or point me towards existing documentation for these parameters, that would be much appreciated.  Thanks!

felixcheung:
unfortunately there isn't much information on SVDPlusPlus - let's skip that one for now

src/main/scala/org/graphframes/lib/ShortestPaths.scala
felixcheung: right - so perhaps not a good idea to duplicate the check here, if Spark DataFrame is already checking it?

SemyonSinchenko:
@rjurney @SauronShepherd I like this PR. It may be slightly incomplete, but it is still a good starting point. What do you think about it? If you are OK with it, I can try to resolve conflicts and we can merge it.

SauronShepherd:
It's not a bad idea, but maybe we should stop for a moment and rethink all these config-related issues/new proposals.

rjurney:
If it helps and makes sense, I like the PR. If we want a more complete approach, we can do that too but let’s not hold up contributions that aren’t sweeping as it precludes most people from contributing.

Issue #159, opened 2016-12-31
Incorrect and non-deterministic connectedComponents() results when the vertices and edges are not cached


nchammas:
I am seeing some surprising and very worrying behavior for `connectedComponents()`. I am running it on a graph with ~300K vertices and ~275K edges. When I run it locally, I get consistent, correct results. When I run it on our cluster, I get inconsistent, incorrect results.

I can't share a repro at the moment, so let me elaborate a bit on how the results are incorrect when I run `connectedComponents()` on our cluster.

When I run `connectedComponents()` on the cluster, about 20-30% of the component IDs returned (so, ~60K-100K) are the same. It's usually a low number from 1 to 10. This is totally incorrect, since I know from running the same code locally and inspecting the results that 200 vertices at most should share the same component ID.

Additionally, when I re-run the same job on our cluster I get different counts every time. So one time it may be ~60K vertices that get assigned a component ID of 0; another time ~80K vertices will be assigned a component ID of 7; and so on. I don't see any such behavior when running locally. When I run this same job locally, I get consistent counts: The most frequently occurring component ID is always shared by exactly 171 vertices.

I don't have any hints about what may be going wrong on the cluster, but occasionally I do see errors like this pop up right around when `connectedComponents()` runs:

```
WARN TransportChannelHandler: Exception in connection from /x.x.x.x:55250
ERROR TransportResponseHandler: Still have 1 requests outstanding when connection from /x.x.x.x:55250 is closed
WARN BlockManagerMaster: Failed to remove broadcast 56 with removeFromMaster = true - Connection reset by peer
```

I don't see them every time though, so they may not be related to the incorrect behavior I'm seeing.

I did figure out a way to work around this problem. If I `persist()` the vertices and edges before calling `connectedComponents()`, I get correct, consistent results like I do when running locally. So **this smells like an optimizer bug in Spark**, loosely related to others like [SPARK-18589](https://issues.apache.org/jira/browse/SPARK-18589) and [SPARK-18492](https://issues.apache.org/jira/browse/SPARK-18492) where adding in a `persist()` at the right place works around the issue.

If someone is interested in tracking this down, I would be happy to work with them to shed more light on what's happening and perhaps boil things down to a reproducible example.

Cluster Environment:
* GraphFrames 0.3.0
* Spark 2.0.2 or Spark 2.1.0
* Scala 2.11
* Java 7
* Python 3.5
* YARN 2.6.0, CDH 5.5.6

Local Environment (same as cluster except for):
[...]

PR #189, opened 2017-05-04
[WIP] [#159] Fix non-deterministic ID assignment

mengxr:
This is a WIP PR that tries to fix nondeterministic vertex ID assignment in GraphFrames (#159). We used to convert vertex DF to RDD and use zipWithUniqueID there. However, if the input DF is not deterministic, the outcome is still nondeterministic. This PR use hash partition + distinct + sort within partition for the ID assignment. Apparently this is more expensive, but hopefully this leads to correct result.

This is the generated plan. It seems correct (not pushing mono_id down).
~~~
scala> spark.range(1000).repartition('id).distinct().sortWithinPartitions('id).withColumn("long_id", monotonically_increasing_id()).explain()

== Physical Plan ==
*Project [id#7L, monotonically_increasing_id() AS long_id#13L]
+- *Sort [id#7L ASC NULLS FIRST], false, 0
   +- *HashAggregate(keys=[id#7L], functions=[])
      +- *HashAggregate(keys=[id#7L], functions=[])
         +- Exchange hashpartitioning(id#7L, 200)
            +- *Range (0, 1000, step=1, splits=Some(4))
~~~

The PR is still WIP. Once we confirm this is the right fix, I will clean up the code and add tests.

@nchammas Could you help test this patch? I don't have data to reproduce the issue. Thanks a lot!

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -466,8 +466,8 @@ class GraphFrame private(
         col(ATTR + "." + ID).cast("long").as(LONG_ID), col(ATTR + "." + ID).as(ID), col(ATTR))
     } else {
       val withLongIds = vertices.select(ID)
+        .repartition(col(ID))
         .distinct()
-        .repartition(1024, col(ID))
         .sortWithinPartitions(ID)
         .withColumn(LONG_ID, monotonically_increasing_id())
         .persist(StorageLevel.MEMORY_AND_DISK)
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -22,9 +22,8 @@ import java.util.Random
 import scala.reflect.runtime.universe.TypeTag
 
 import org.apache.spark.graphx.{Edge, Graph}
-import org.apache.spark.sql.SQLHelpers._
 import org.apache.spark.sql._
-import org.apache.spark.sql.functions.{array, broadcast, col, count, explode, struct, udf}
+import org.apache.spark.sql.functions.{array, broadcast, col, count, explode, struct, udf, monotonically_increasing_id}
 import org.apache.spark.sql.types._
 import org.apache.spark.storage.StorageLevel
 
@@ -466,8 +465,15 @@ class GraphFrame private(
       indexedVertices.select(
         col(ATTR + "." + ID).cast("long").as(LONG_ID), col(ATTR + "." + ID).as(ID), col(ATTR))
     } else {
-      val indexedVertices = zipWithUniqueId(vertices)
-      indexedVertices.select(col("uniq_id").as(LONG_ID), col("row." + ID).as(ID), col("row").as(ATTR))
+      val withLongIds = vertices.select(ID)
+        .distinct()
+        .repartition(1024, col(ID))
+        .sortWithinPartitions(ID)
+        .withColumn(LONG_ID, monotonically_increasing_id())
+        .persist(StorageLevel.MEMORY_AND_DISK)
+      vertices.select(col(ID), nestAsCol(vertices, ATTR))
+        .join(withLongIds, ID)
+        .select(LONG_ID, ID, ATTR)
     }
   }
 


nchammas:
@mengxr - I'll try to test this tomorrow or early next week.

After checking out this branch and building per the README instructions, how do I get this loaded on our YARN cluster and referenced by `spark-submit` for testing? Do I just copy the JAR to some spot on HDFS and reference it by path in the call to `spark-submit --packages ...`?

I don't see #159 when running locally, so I need to test this on our cluster.

mengxr:
You can try a local standalone cluster, which should give the same result as on YARN. Or you can build the graphframes jar by `sbt -DskipTest assembly`, and then use `--jars` to include it in your job. `--packages` only fetches the official release.

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -466,8 +465,15 @@ class GraphFrame private(
       indexedVertices.select(
         col(ATTR + "." + ID).cast("long").as(LONG_ID), col(ATTR + "." + ID).as(ID), col(ATTR))
     } else {
-      val indexedVertices = zipWithUniqueId(vertices)
-      indexedVertices.select(col("uniq_id").as(LONG_ID), col("row." + ID).as(ID), col("row").as(ATTR))
+      val withLongIds = vertices.select(ID)
+        .distinct()
```
jkbradley: Does distinct() do sorting or hashing already?  Could you hash partition and then implement distinct() within each partition?  I'm not sure how smart Catalyst is here.

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -466,8 +465,15 @@ class GraphFrame private(
...
+        .repartition(1024, col(ID))
```
jkbradley: Can numPartitions be set according to the original DataFrame numPartitions?

src/main/scala/org/graphframes/GraphFrame.scala
mengxr: Yes, but it doesn't specify which method it uses and it doesn't allow users to set a partition number. Based on the optimized plan, it does shuffle twice. I just wanted to put a quick fix for @nchammas to test whether this is the root cause.

src/main/scala/org/graphframes/GraphFrame.scala
mengxr: That would be very expensive if the original DataFrame has few partitions. Again, this is just a quick patch to see whether it fixes the problem. We need more work after confirmation.

mengxr:
@nchammas Did you get a chance to test the patch?

mengxr:
It is hard to reproduce the nondeterministic behavior on local. But it is fairly easy to show distinct doesn't guarantee ordering on a multi-node cluster:

screen shot 2017-05-08 at 8 56 35 am

nchammas:
@mengxr - Will try today.

nchammas:
@mengxr - I removed `--packages` and instead included the locally built `graphframes-assembly-0.4.0-SNAPSHOT-spark2.1.jar` via `--jars`. However, Spark can't find GraphFrames:

```
ImportError: No module named 'graphframes'
```

I built GraphFrames using `sbt "set test in assembly := {}" assembly` from your comment above. Is there anything else I need to do?

mengxr:
Did you use python or scala?

nchammas:
My app is Python.

mengxr:
Then you also need to include the python files, which should be included in the jar. Could you test it in Python instead?

nchammas:
So I need to include `python/graphframes/graphframe.py` as well, I guess? Or should I zip up all of `python/graphframes` and include that? Probably the latter.

> Could you test it in Python instead?

I presume you meant Scala here, but my app is in Python and I'd like to test this with my app to be sure.

mengxr:
Yes, you need the python file, which should be bundled in the assembly jar. But you need to put the jar on the `PYTHONPATH`. So please try to add it to both `--jars` and `--py-files`:

~~~
--jars target/scala-2.11/graphframes-assembly-0.4.0-SNAPSHOT-spark2.1.jar --py-files target/scala-2.11/graphframes-assembly-0.4.0-SNAPSHOT-spark2.1.jar
~~~

nchammas:
OK adding the JAR to `--py-files` worked, thanks.

Looks like GraphFrames @ 2c89c6d967c1c2a8d7ba1e02b16fee1ad99a9a1a fixes the problem for me. 👏 Do I need to test again @ 398d8c6c22c1bc025e0f749a5ee6aea4ba1c9a56?

mengxr:
Not necessary, it is just an optimization. Thanks for confirming the fix! I will update the PR to make it looks better:) And we will probably make a bug fix release after since this is critical.

mengxr:
@phi-dbq will take over the follow-up tasks.

mengxr:
I'm closing this PR for now. @phi-dbq will take the follow-up work in #159.

nchammas:
> @phi-dbq will take the follow-up work in #159.

I think you mean #195.

Issue #184, opened 2017-04-19
Add wrapper for GraphX parallelized personalized PageRank


jkbradley:
GraphX added parallelized personalized PageRank in Spark 2.1.  It'd be nice to add a wrapper for that in GraphFrames.  If you're interested in working on this, please state the proposed API before you begin implementation.  Thanks!

jkbradley:
@phi-dbq is interested in taking this.  Thanks!

jkbradley:
Scala API merged.  I'll leave this issue open for the Python API + user guide updates.

jkbradley:
All done!

PR #191, opened 2017-05-09
[184] Python API for Parallel Personalized PageRank

phi-dbq:
This is a PR that adds the Python API for GraphFrames (#184).

### DIFF
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -19,6 +19,8 @@ import sys
 if sys.version > '3':
     basestring = str
 
+from py4j.java_gateway import JavaGateway
+
 from pyspark import SparkContext
 from pyspark.sql import Column, DataFrame, SQLContext
 from pyspark.storagelevel import StorageLevel
@@ -333,12 +335,18 @@ class GraphFrame(object):
         :param maxIter: the fixed number of iterations this algorithm runs
         :return:  GraphFrame with new vertices column "pageranks" and new edges column "weight"
         """
-        assert sourceIds is not None, "Source vertices Ids sourceIds must be provided"
+        assert sourceIds is not None and len(sourceIds) > 0, "Source vertices Ids sourceIds must be provided"
         assert maxIter is not None, "Max number of iterations maxIter must be provided"
+        if isinstance(sourceIds, list):
+            _gateway = self._sc._gateway
+            _srcIds = _gateway.new_array(_gateway.jvm.Object, len(sourceIds))
+            for i, vid in enumerate(sourceIds):
+                _srcIds[i] = vid
+            sourceIds = _srcIds
         builder = self._jvm_graph.parallelPersonalizedPageRank()
-        builder.resetProbability(resetProbability)
-        builder.sourceIds(sourceIds)
-        builder.maxIter(maxIter)
+        builder = builder.resetProbability(resetProbability)
+        builder = builder.sourceIds(sourceIds)
+        builder = builder.maxIter(maxIter)
         jgf = builder.run()
         return _from_java_gf(jgf, self._sqlContext)
 
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -320,6 +320,28 @@ class GraphFrame(object):
         jgf = builder.run()
         return _from_java_gf(jgf, self._sqlContext)
 
+    def parallelPersonalizedPageRank(self, resetProbability = 0.15, sourceIds = None,
+                                     maxIter = None):
+        """
+        Run the personalized PageRank algorithm on the graph,
+        from the provided list of sources in parallel for a fixed number of iterations.
+
+        See Scala documentation for more details.
+
+        :param resetProbability: Probability of resetting to a random vertex
+        :param sourceIds: the source vertices for a personalized PageRank
+        :param maxIter: the fixed number of iterations this algorithm runs
+        :return:  GraphFrame with new vertices column "pageranks" and new edges column "weight"
+        """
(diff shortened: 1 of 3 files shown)


python/graphframes/graphframe.py
```diff
@@ -320,12 +322,18 @@ def parallelPersonalizedPageRank(self, resetProbability = 0.15, sourceIds = None
         :param maxIter: the fixed number of iterations this algorithm runs
         :return:  GraphFrame with new vertices column "pageranks" and new edges column "weight"
         """
-        assert sourceIds is not None, "Source vertices Ids sourceIds must be provided"
+        assert sourceIds is not None and len(sourceIds) > 0, "Source vertices Ids sourceIds must be provided"
         assert maxIter is not None, "Max number of iterations maxIter must be provided"
+        if isinstance(sourceIds, list):
+            _gateway = self._sc._gateway
+            _srcIds = _gateway.new_array(_gateway.jvm.Object, len(sourceIds))
```
phi-dbq: The `Scala API` expects `sourceIds` to be of type `Array[Any]`. 
Directly passing a `Python list` triggers an error. 
We explicitly create a `Java Array` and copy the elements to it.

python/requirements.txt
```diff
@@ -1,2 +1,3 @@
 # This file should list any python package dependencies.
-nose==1.3.3
\ No newline at end of file
+nose==1.3.3
+numpy>=1.7
```
phi-dbq: need `numpy` for `spark.ml`

python/graphframes/graphframe.py
```diff
@@ -307,6 +309,34 @@ def pageRank(self, resetProbability = 0.15, sourceId = None, maxIter = None,
...
+        :param resetProbability: Probability of resetting to a random vertex
+        :param sourceIds: the source vertices for a personalized PageRank
+        :param maxIter: the fixed number of iterations this algorithm runs
+        :return:  GraphFrame with new vertices column "pageranks" and new edges column "weight"
+        """
+        assert sourceIds is not None and len(sourceIds) > 0, "Source vertices Ids sourceIds must be provided"
+        assert maxIter is not None, "Max number of iterations maxIter must be provided"
+        if isinstance(sourceIds, list):
+            _gateway = self._sc._gateway
+            _srcIds = _gateway.new_array(_gateway.jvm.Object, len(sourceIds))
```
phi-dbq: We have to manually convert `Python list` to `Java Array[Any]`. Not sure if there is a better way.

phi-dbq:
@jkbradley

python/graphframes/graphframe.py
felixcheung: like here?https://github.com/apache/spark/blob/a8ea4da8d04c1ed621a96668118f20739145edd2/core/src/main/scala/org/apache/spark/api/python/PythonUtils.scala#L68

python/graphframes/graphframe.py
```diff
@@ -337,12 +335,7 @@ def parallelPersonalizedPageRank(self, resetProbability = 0.15, sourceIds = None
         """
         assert sourceIds is not None and len(sourceIds) > 0, "Source vertices Ids sourceIds must be provided"
         assert maxIter is not None, "Max number of iterations maxIter must be provided"
-        if isinstance(sourceIds, list):
-            _gateway = self._sc._gateway
-            _srcIds = _gateway.new_array(_gateway.jvm.Object, len(sourceIds))
-            for i, vid in enumerate(sourceIds):
-                _srcIds[i] = vid
-            sourceIds = _srcIds
+        sourceIds = self._sc._jvm.PythonUtils.toArray(sourceIds)
```
phi-dbq: Using the `PythonUtils.toArray` method to convert `Python list` type. Thanks @felixcheung !

phi-dbq:
Any additional comments? @jkbradley @felixcheung

python/graphframes/tests.py
```diff
@@ -227,6 +261,18 @@ def test_page_rank(self):
         pr = g.pageRank(resetProb, tol=errorTol)
         self._hasCols(pr, vcols=['id', 'pagerank'], ecols=['src', 'dst', 'weight'])
 
+    def test_parallel_personalized_page_rank(self):
+        if GraphFrameTestUtils.is_later_version("2.1"):
```
felixcheung: is it better to do `@unittest.skipIf` https://docs.python.org/2/library/unittest.html
this way it's indicated the test is skipped?

python/graphframes/tests.py
```diff
@@ -35,6 +36,33 @@
...
 
+class GraphFrameTestUtils(object):
+
+    @classmethod
+    def parse_spark_version(cls, version_str):
+        _sc_ver_patt = r'(\d+)\.(\d+)(\.(\d+)(-(.+))?)?'
+        m = re.match(_sc_ver_patt, version_str)
+        _major_version = m.group(1)
+        _minor_version = m.group(2)
+        return _major_version, _minor_version
```
felixcheung: should we also have patch version too?

python/graphframes/tests.py
phi-dbq: It is a bit tricky as determining current spark version requires the `sc` object. When using if with `@unittest.skipIf` annotation, an error is reported. It looks like `sc` is not initialized when the annotation is evaluated.

felixcheung: I see - we could define an environment variable like we have for SCALA_VERSION

https://github.com/graphframes/graphframes/blob/master/python/run-tests.sh#L47

but agreed it might not be worthwhile for the effort

felixcheung: actually, instead of a decorator, this can be called in code
https://docs.python.org/2/library/unittest.html#unittest.skipIf

PR #192, opened 2017-05-10
[WIP][159] Fix non-deterministic ID assignment

phi-dbq:
This is a follow up task from [#189]. 
It removes `SQLHelpers.zipWithUniqueId` which is no longer needed.
We will make scalability test and address potential issues.
In the end, we will make a bug-fix release based on changes in this PR.


jkbradley:
@phi-dbq What about removing the 2.0 shim?

PR #196, opened 2017-05-14
[57] [170] aggregateMessages for python update

tsivula:
This PR implements the updates discussed in PR #169 post merge.

Updates:
- Move AggregateMessages under graphframes.lib
- rename `msgToX` into `sendToX`
- change parameter ``assert`` checks into raise errors
- check for string type (Python 2 and 3) for params `sendToX`

### DIFF
--- a/python/graphframes/lib/aggregate_messages.py
+++ b/python/graphframes/lib/aggregate_messages.py
@@ -1,0 +1,83 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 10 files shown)


python/graphframes/lib/aggregate_messages.py
```diff
@@ -0,0 +1,83 @@
...
+#
+
+from pyspark import SparkContext
+from pyspark.sql import DataFrame, functions as sqlfunctions
+
+
+def _java_api(jsc):
+    javaClassName = "org.graphframes.GraphFramePythonAPI"
+    return jsc._jvm.Thread.currentThread().getContextClassLoader().loadClass(javaClassName) \
+            .newInstance()
```
felixcheung: interesting.. this wasn't there in the earlier implementation?

felixcheung:
@jkbradley for reviewing this and heads up before rolling another release.

python/graphframes/lib/aggregate_messages.py
tsivula: This is a copy from https://github.com/graphframes/graphframes/pull/196/files/61c9ae6fc942a47d76735e040d274f6484fab08c#diff-e81e6b169c0aa35012a3263b2f31b330L32
... from the module this new module was detached from. Maybe we should just import it instead?
``from graphframes.graphframe import _java_api``

jkbradley: Importing sounds reasonable

jkbradley:
@tsivula Thanks for the update!
@felixcheung This LGTM, so I'll go ahead and merge it.

python/graphframes/lib/aggregate_messages.py
jkbradley: Whoops, just merged this.  We can do the import sometime later.  I'll go ahead and start preparing a release to include this and the indexing bug fix.

tsivula:
Thanks for the reviewers

PR #197, opened 2017-05-16
[184][Docs] Parallel Personalized PageRank docs update

phi-dbq:
This is a follow up `PR` to update the documentation of the `ParallelPersonalizedPageRank` [#184].
Docs are built following `docs/README.md`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,6 +27,11 @@ in some cases:
     $ sudo gem install jekyll
     $ sudo gem install jekyll-redirect-from
 
+On macOS, with the default Ruby, please install Jekyll with Bundler as [instructed on offical website](https://jekyllrb.com/docs/quickstart/). Otherwise the build script might fail to resolve dependencies.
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -27,6 +27,11 @@ in some cases:
     $ sudo gem install jekyll
     $ sudo gem install jekyll-redirect-from
 
+On `macOS`, with the default ruby,
+
+    $ sudo gem install jekyll bundler
+	$ sudo gem install jekyll-redirect-from
```
felixcheung: indentation is a bit strange here?

felixcheung: any why list them when they are the same as L27, L28?

phi-dbq: 1. Sorry about the indentation issue - forgot to `untabify`. 
2. It seems that without using the `bundler` version, `jekyll` fails to find `pygments`.

Issue #203, opened 2017-06-06
Update ConnectedComponent.symmetrize once we remove Spark 1.6 support


mengxr:
The current implementation (https://github.com/graphframes/graphframes/blob/master/src/main/scala/org/graphframes/lib/ConnectedComponents.scala#L170) requires two passes over the edges, mainly to have the same code to support both Spark 1.6 and 2.x. Once we removed Spark 1.6 support (#193), we should optimize the implementation to speed-up CC.

felixcheung:
closed by #205

Issue #193, opened 2017-05-10
Remove Apache Spark 1.6 support


jkbradley:
I'd like to remove 1.6 support
* I have not heard any complaints about not publishing a 1.6 version of the latest GraphFrames release.
* It is causing complications for https://github.com/graphframes/graphframes/issues/159  (where tests are suddenly much slower, though I'm not sure why yet).
* Later version of Spark are consistently better for GraphFrames b/c of DataFrame performance improvements

jkbradley:
CC @nchammas @felixcheung  Does this sound OK to you?

nchammas:
Fine with me. We use GraphFrames with 2.0.2 and are planning to upgrade to 2.1.1 this month.

felixcheung:
Ok, I checked around. Probably much easier to maintain if only on 2.x release

jkbradley:
OK thanks!  I'll send a PR soon

felixcheung:
closed by #204

PR #205, opened 2017-06-10
[#203] optimize cc.symmetrize for Spark 2.0+

mengxr:
This PR uses a single pass to create bi-directional edges in CC. This only works after we remove 1.6 support (#193, #204). We should see significant speed-up in travis builds.


mengxr:
Seems that this cuts the build time by more than half (23 mins -> 9 mins).

mengxr:
The failed 1.6 builds are expected.

PR #207, opened 2017-06-12
Set project name explicitly in build.sbt

mengxr:
I had several local folders to test different patches. Without explicitly setting the project name, the assembly jar names end up differently, which affects python tests. This PR sets the name explicitly, so we can rename the root folder freely.


mengxr:
merging this into master

PR #208, opened 2017-06-13
ConnectedComponents: test differing edge directions

blendmaster:
I ran into this porting some graphx code to graphframes in some of my own unit tests. I dunno what's going on yet, but it appears to be minimal, after trying to shrink the test case manually.


blendmaster:
In case it wasn't obvious, I was expecting a test failure:

```
[info] - one component, differing edge directions *** FAILED ***

[info]   Set(Set(0, 1, 2, 4), Set(2, 3)) did not equal Set(Set(0, 1, 2, 3, 4)) (ConnectedComponentsSuite.scala:232
```

mengxr:
Seems like a bad bug ... I will take a look.

mengxr:
I think I found the bug. The min neighbor could be the vertex itself. Will submit a fix later today.

mengxr:
This was fixed by #212.

PR #212, opened 2017-06-13
[208] Fix cc min nbr in small star

mengxr:
When a vertex (e.g., 1) is the smallest among its neighbors (e.g., 2, 3, 4), we didn't exclude the record `(1, 2)` from the output of `minNbrs`. The bug was reported by @blendmaster in #208.

Closes #208


jkbradley:
This LGTM, comparing with the Kiveris et al (2014) paper, but it'd be great to get a 2nd review if @felixcheung or others have the time

thunterdb:
LGTM

PR #215, opened 2017-06-14
Make intermediate storage level configurable in CC

mengxr:
The default storage level `MEMORY_AND_DISK` might not be the best option for certain setup. We can make it configurable by users.

* The test only checks the impl works under different settings but it doesn't verify the requested storage level is actually used.
* Python API can be added in a separate PR.


phi-dbq:
LGTM!

Issue #219, opened 2017-06-27
Make edgeStorageLevel and vertexStorageLevel configurable


estebandonato:
Most of the GraphFrame algorithms are wrappers of GraphX algorithm implementation. That's the case of PageRank, LabelPropagation, StronglyConnectedComponents, etc. It turns out that all these algorithms use the Pregel api implemented in GraphX. Internally, Pregel caches the vertex and edge RDDs on each iteration. The storage levels used by Pregel to cache the vertices and edges are the ones passed to Graph.apply method when a graphX instance is created. The default storage level for both vertex and edge RDDs is MEMORY_ONLY as we can see it in Spark code:

```
def apply[VD: ClassTag, ED: ClassTag](
      vertices: RDD[(VertexId, VD)],
      edges: RDD[Edge[ED]],
      defaultVertexAttr: VD = null.asInstanceOf[VD],
      edgeStorageLevel: StorageLevel = StorageLevel.MEMORY_ONLY,
      vertexStorageLevel: StorageLevel = StorageLevel.MEMORY_ONLY): Graph[VD, ED]
```

GraphFrames algorithms convert a GraphFrame instance to a GraphX instance before calling to the graphX algorithms but using the default Storage Level. This might not be a good choice in some environments of limited RAM.

My proposal is to make both edgeStorageLevel and vertexStorageLevel configurable per algorithm. Internally this would create a graphX instance with the configured storage levels before passing the instance to Pregel. This is somehow similar to feature #213 but for all the algorithms that are wrappers of a graphX implementation.

We are needing this feature so if you give me green light I can work on it.

thunterdb:
@estebandonato that sounds like a great idea. Feel free to submit a PR along the lines of what was done in #213 .

felixcheung:
do you have a specific need for vertex and edge at different storage level?

estebandonato:
@felixcheung no, I usually use the same storage level for both vertex and edge, however I would like to leave the api open to contemplate any combination. Thoughts?

estebandonato:
Finally I have finished PR #225 which addresses this issue. Please review it and let me know your feedback.

estebandonato:
@thunterdb @felixcheung I was wondering if any of you guys could review PR #225 that addresses this issue and provide feedback. Thanks!

felixcheung:
yes! will do shortly. thanks for the patience.

Issue #213, opened 2017-06-13
Allow users to set intermediate storage level in CC


mengxr:
This would help users test different storage settings for performance tuning.

mengxr:
done by #215

PR #225, opened 2017-08-02
Make edgeStorageLevel and vertexStorageLevel configurable

estebandonato:
This PR addresses issue #219 The properties ``intermediateVertexStorageLevel`` and ``intermediateEdgeStorageLevel`` with their respective getters and setters were added to the classes that represent graph algorithms with a GraphX implementation. These 2 properties are used to change the storage level of the GraphX instance referenced by ``GraphFrame.cachedTopologyGraphX`` before calling the GraphX algorithm implementation.

In the special case of ConnectedComponent class, the properties were named ``graphxVertexStorageLevel`` and ``graphxEdgeStorageLevel`` to differentiate them with the property ``intermediateStorageLevel`` used in the graphframe implementation of this algorithm. Also the Python api of ConnectedComponent was modified to include a similar feature implemented in #213 only to the Scala api

These changes were applied to both the Scala and Python apis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 jdk:
-  - oraclejdk7 # openJDK crashes sometimes
+  - openjdk7 # openJDK crashes sometimes
 
 sudo: required
 
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -259,7 +259,10 @@ class GraphFrame(object):
     # Standard algorithms
 
     def connectedComponents(self, algorithm = "graphframes", checkpointInterval = 2,
-                            broadcastThreshold = 1000000):
+                            broadcastThreshold = 1000000, 
+                            intermediateStorageLevel = StorageLevel.MEMORY_AND_DISK,
+                            intermediateGraphxVertexStorageLevel = StorageLevel.MEMORY_ONLY,
+                            intermediateGraphxEdgeStorageLevel = StorageLevel.MEMORY_ONLY):
         """
         Computes the connected components of the graph.
 
@@ -270,6 +273,14 @@ class GraphFrame(object):
         :param checkpointInterval: checkpoint interval in terms of number of iterations (default: 2)
         :param broadcastThreshold: broadcast threshold in propagating component assignments
           (default: 1000000)
+        :param intermediateStorageLevel: storage level for intermediate datasets that require 
+               multiple passes (default: ``MEMORY_AND_DISK``)  
+        :param intermediateGraphxVertexStorageLevel: storage level for intermediate `Graph` vertices that 
+               require multiple passes (default: `MEMORY_ONLY`). This parameter is only used when the 
+               algorithm is set to "graphx"
+        :param intermediateGraphxEdgeStorageLevel: storage level for intermediate `Graph` edges that 
+               require multiple passes (default: `MEMORY_ONLY`). This parameter is only used when the 
+               algorithm is set to "graphx"
 
         :return: DataFrame with new vertices column "component"
         """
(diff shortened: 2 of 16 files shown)


estebandonato:
oraclejdk7 is no longer available in Trusty, so I modified travis configuration to use openjdk7. More details https://github.com/travis-ci/travis-ci/issues/7964

I'm done with this PR and waiting for your feedback

felixcheung:
instead of having these in each GF methods, I wonder if this should be a property of the "Graph" or "GraphFrame"?

.travis.yml
```diff
@@ -1,5 +1,5 @@
 jdk:
-  - oraclejdk7 # openJDK crashes sometimes
+  - openjdk7 # openJDK crashes sometimes
```
felixcheung: can you rebase to pick up the latest?

estebandonato: done!

estebandonato:
@felixcheung actually I thought implementing this feature the way you describe, however I followed the current approach mainly to be aligned with the changes in api implemented in #213 and to allow different storage level settings for different algorithm executions. On the other hand, if we implement vertex and edge storage levels as GraphFrame properties we would be aligned with the approach used in GraphX. Both options have their pros and cons. Just let me know what you guys think and I can change the code if we decide to go with the properties approach.

felixcheung:
right, I'm just worry the impact of having to have 2 parameters for every one graph algo we have.
let's open this up and see anyone has any preference on the approach?

mengxr:
@estebandonato Two comments:

1) Why do you need to control the storage levels for vertices/edges and in GraphX? I'm curious about the scenarios.

2) To prevent adding too many parameters to APIs, we can introduce a new class and predefined constants:

~~~scala
class GraphStorageLevel(vertexStorageLevel, edgeStorageLevel)

object GraphStorageLevel {
  val MEMORY_ONLY = ...
  val ...
}
~~~

Then in each method we only need one arg to describe the storage levels. We can overload the current method to take this type. But this is only needed if 1) can be justified.

estebandonato:
@mengxr regarding your question 1) we need to control storage levels in GraphX mainly because most of the GraphFrame's graph algorithms are just wrappers of the GraphX implementations. These are the cases of PageRank, LabelPropagation, StronglyConnectedComponents, just to list some of them. These are all iterative algorithms which extensively cache vertices and edges on each iteration to avoid re-computation. Concretely, that's the case of Pregel implementation in GraphX https://github.com/apache/spark/blob/v2.1.1/graphx/src/main/scala/org/apache/spark/graphx/Pregel.scala#L112. Storage levels used when caching a GraphX instance are defined in the Graph.apply method as it is shown in the Spark code below

```
def apply[VD: ClassTag, ED: ClassTag](
      vertices: RDD[(VertexId, VD)],
      edges: RDD[Edge[ED]],
      defaultVertexAttr: VD = null.asInstanceOf[VD],
      edgeStorageLevel: StorageLevel = StorageLevel.MEMORY_ONLY,
      vertexStorageLevel: StorageLevel = StorageLevel.MEMORY_ONLY): Graph[VD, ED]
```
Currently, when GraphFrame has to call a GraphX algorithm, it first converts a GraphFrame instance to a GraphX instance but there is no way to specify vertex & edge storage levels when creating the GraphX instance. As consequence the default ``MEMORY_ONLY`` storage level is used to cache the graph on each iteration. In same environments of limited memory, parts that cannot fit in memory are not cached and so they are recomputed, and the cost of recomputing parts gets worse on each new iteration. Specifically what we have experienced, for instance with PageRank is that performance degraded on each iteration. We have also seen that the amount of shuffle reads for the same tasks increased on each iteration. As workaround we had to copy the wrapper code implemented in GraphFrame in our code but changing the way a GraphFrame instance is converted to a GraphX instance so we can set the storage levels. With that workaround the response times and shuffle reads stopped degrading on each iteration. Hope this justify the change. Eventually if there are plans to migrate the graphX implementation to a custom GraphFrame implementation (like the case of connected component) we could re-use these parameters for the new implementation.

Regarding 2) if we want to prevent adding too many parameters to the API, another alternative is just to have 1 storage level value per algorithm and apply it to both vertices and edges

Let me know your feedback to make the changes accordingly.

estebandonato:
@mengxr did you have the chance to read the reasons of this change explained above? Please let me know your thoughts

estebandonato:
@mengxr any update on this?

SemyonSinchenko:
Closed in favor of #698

Issue #233, opened 2017-09-14
Inaccurate PageRank docs


danielfx90:
## **Current situation**
PageRank scala docs and user guide say that the implementation which runs for a fixed number of iterations (through the `maxIter` parameter) uses the Graphframes interface when in fact it uses the graphX's Pregel one.

## **What needs to be done?**
Both docs should match what the code does.

**Note:** I'll be creating a PR for this

felixcheung:
fixed by 234

PR #234, opened 2017-09-14
PageRank docs updated

danielfx90:
A solution to issue #233

## **Current situation**
PageRank scala docs and user guide say that the implementation which runs for a fixed number of iterations (through the `maxIter` parameter) uses the Graphframes interface when in fact it uses the graphX's Pregel one.

## **What was done?**
Both docs were fixed to match what the code does.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -642,12 +642,12 @@ result.select("id", "label").show()
 
 ## PageRank
 
+There are two implementations of PageRank.
 
-There are two implementations of PageRank. Both of them use the `org.apache.spark.graphx.Pregel` interface,
-however, they differ on the stopping condition.
-
-* The first one runs PageRank for a fixed number of iterations and this can be run by setting `maxIter`.
-* The second implementation runs PageRank until convergence and this can be run by setting `tol`.
+* The first one uses the `org.apache.spark.graphx.graph` interface with `aggregateMessages` and runs
+PageRank for a fixed number of iterations. This can be executed by setting `maxIter`.
+* The second implementation uses the `org.apache.spark.graphx.Pregel` interface and runs PageRank until
+convergence and this can be run by setting `tol`.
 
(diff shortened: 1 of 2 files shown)


felixcheung:
hi! thanks for the contribution. GraphFrame is definitely calling GraphX - but GraphX's PageRank implementation is done on `aggregateMessages` and not on `Pregel`?

danielfx90:
You are right! Let me fix that

danielfx90:
@felixcheung All done! Thanks for the correction. Meanwhile, I realized that all current PRs are failing Travis checks because it can not find the `pip3` command:
https://travis-ci.org/graphframes/graphframes/jobs/275643520#L573

I think that bug got leaked into `master`

felixcheung:
I've merged in the fix, please rebase

danielfx90:
@felixcheung done!

docs/user-guide.md
```diff
@@ -644,10 +644,10 @@ result.select("id", "label").show()
...
 There are two implementations of PageRank.
 
-* The first implementation uses the standalone `GraphFrame` interface and runs PageRank
- for a fixed number of iterations.  This can be run by setting `maxIter`.
-* The second implementation uses the `org.apache.spark.graphx.Pregel` interface and runs PageRank
-  until convergence.  This can be run by setting `tol`.
+* The first one uses the `org.apache.spark.graphx.graph` interface with `aggregateMessages` and runs
+PageRank for a fixed number of iterations. This can be executed by setting `maxIter`.
+* The second implementation uses the `org.apache.spark.graphx.Pregel` interface and runs PageRank until
+convergence and this can be run by setting `tol`.
```
felixcheung: was there a preference between `run` and `executed` (above)?

danielfx90: Only to avoid an overuse of the verb `run`. Both words seem to fit the action described, but let me know if it feels too awkward.

felixcheung:
merged. thanks

Issue #235, opened 2017-09-15
Personalized PageRank broken in Spark 2.2+


felixcheung:
A change in personalized page rank has broken GraphFrames due to GF's use of Long vertex ID apache/spark#16483

felixcheung:
from work to support 2.2 #223

MrBago:
Was this fixed by #236? Can we close it?

felixcheung:
PR #236 is not merged.

nmurphycb:
Are there plans to merge it in the future?

felixcheung:
I think there are some discussions/concerns on that PR.

Bykow:
I need to compute a personalized pagerank, it is ok to use version 2.1 ? Is there any risks/side effects I could encounter ?

dmvieira:
news?

felixcheung:
awesome

felixcheung:
I guess the answer/fix is: Spark 2.4.0+

Issue #232, opened 2017-09-12
pageRank performance degradation 12 times! in 0.5


artem-aliev:
pageRank run 12 times slower with 0.5 compared to 0.4. What could be a reason?
Is other graph algorithms affected?
```
def time[A](f: => A) = {val s = System.nanoTime; val ret = f; println("time: "+(System.nanoTime-s)/1e6+"ms"); ret }

import org.graphframes._
val g = examples.Graphs.friends
val results2 = time { g.pageRank.resetProbability(0.15).maxIter(10).run() }
```

Spark 2.0.2 graphframes:graphframes:0.4.0-spark2.0-s_2.11, time: 6934ms
Spark 2.0.2 graphframes:graphframes:0.5.0-spark2.0-s_2.11, time: 87887ms

artem-aliev:
It is regression caused by #159 fix.
Removing persist from following line restore performance and should not affect id consistency if vertices are reordered:
https://github.com/graphframes/graphframes/blob/d43779712f586c664e88565dfbaa77a700e186cf/src/main/scala/org/graphframes/GraphFrame.scala#L483

1. I'm surprised that this persist cause a lot of writes. 10MB per stage for small 7 vertex graph:
```
Stage Id  ▾ | Description | Submitted | Duration | Tasks: Succeeded/Total | Input | Output | Shuffle Read | Shuffle Write
-- | -- | -- | -- | -- | -- | -- | -- | --
148 | foreachPartition at PageRank.scala:157+details | 2017/09/14 16:25:07 | 1.0 s | 200/200 | 436.3 KB |   | 10.1 MB |  
147 | mapPartitions at VertexRDDImpl.scala:247+details | 2017/09/14 16:25:02 | 5 s | 200/200 | 14.9 MB |   | 348.0 B | 10.1 MB
146 | mapPartitions at GraphImpl.scala:207+details | 2017/09/14 16:25:02 | 0.1 s | 200/200 | 436.3 KB |   |   | 348.0 B
```

without persist:
```
Stage Id  ▾ | Description | Submitted | Duration | Tasks: Succeeded/Total | Input | Output | Shuffle Read | Shuffle Write
-- | -- | -- | -- | -- | -- | -- | -- | --
151 | foreachPartition at PageRank.scala:157+details | 2017/09/14 16:11:18 | 23 ms | 8/8 | 17.9 KB |   | 14.6 KB |  
[...]

PR #236, opened 2017-09-15
[#232] Fix partition number for stored id mapping for better performance

artem-aliev:
repartition() method by default create 200 or fixed spark.sql.shuffle.partitions partitions
that causes performance problems for small graphs and OOM problems for large ones.
to fix both problem let's make id mapping DF have the same number of partitions that original vertex DF.


src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -476,7 +476,7 @@ class GraphFrame private(
         col(ATTR + "." + ID).cast("long").as(LONG_ID), col(ATTR + "." + ID).as(ID), col(ATTR))
     } else {
       val withLongIds = vertices.select(ID)
-        .repartition(col(ID))
+        .repartition(vertices.rdd.getNumPartitions,col(ID))
```
felixcheung: please add a space after `,`

felixcheung:
actually looks like a different problem with build, let me investigate

felixcheung:
I've merged in the fix, please rebase

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -475,13 +475,13 @@ class GraphFrame private(
...
-      val withLongIds = vertices.select(ID)
-        .repartition(col(ID))
+      val repartitionedVertices = vertices.repartition(vertices.rdd.getNumPartitions, col(ID))
+      val withLongIds = repartitionedVertices.select(ID)
         .distinct()
         .sortWithinPartitions(ID)
         .withColumn(LONG_ID, monotonically_increasing_id())
         .persist(StorageLevel.MEMORY_AND_DISK)
-      vertices.select(col(ID), nestAsCol(vertices, ATTR))
+      repartitionedVertices.select(col(ID), nestAsCol(repartitionedVertices, ATTR))
```
artem-aliev: I played with Cassandra data source and with hdfs bas dataset, and it looks like this join repartition back to 200 partitions the resulted DF. so I explicitly repartition both parts of the join tot the same number of partitions.
I use the "ID" for partitioning both sides, so join should not reparation both of them.

artem-aliev:
I'm sorry for delay with answers. Here is a better fix. It also fix #235 issue.
the performance test results are the same and strongly depends on number of partitions in the provided graph.
```
def time[A](f: => A) = {val s = System.nanoTime; val ret = f; println("time: "+(System.nanoTime-s)/1e6+"ms"); ret }

import org.graphframes._
val g = examples.Graphs.friends
val results2 = time { g.pageRank.resetProbability(0.15).maxIter(10).run() }
```

it is now: 72273ms
with fix:  9996ms

if i coalesce the graph I got even better result with the fix:
```
val g2 = GraphFrame (g.vertices.coalesce(1), g.edges.coalesce(1))
val results2 = time { g.pageRank.resetProbability(0.15).maxIter(10).run() }
```
7996ms

artem-aliev:
Do you mean add more unit tests? Give me some time, I'll try to create some...

artem-aliev:
Sorry for the long delay. While I doing other thing I realised that
DF monotonically_increasing_id() method assign id that are greater them max int,
It has paritition id in higher 32 bits, so we got #235 issue. The algorithm require id to be less then max int.
RDD has zipWithIndex method that provide proper indexing, DF has no one, so I convert DF to rdd add index and convert  back. Performance is still as good as it was as most time is spend in graph algorithm iterations.
The test that check correctnes of id assignment is aslo added. It aslo checks that id < Int.maxValue
Is it ok or I did not know other pit falls?

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -475,13 +474,17 @@ class GraphFrame private(
...
-        .repartition(col(ID))
-        .distinct()
-        .sortWithinPartitions(ID)
-        .withColumn(LONG_ID, monotonically_increasing_id())
-        .persist(StorageLevel.MEMORY_AND_DISK)
-      vertices.select(col(ID), nestAsCol(vertices, ATTR))
+      val repartitionedVertices = vertices.repartition(vertices.rdd.getNumPartitions, col(ID))
+      val idDF = repartitionedVertices.select(ID)
+      val idType = idDF.schema(0)
+      // DF monotonically_increasing_id() method assign id that are greater them max int,
```
amuraru: @artem-aliev 
zipWithIndex generates a Long index  isn't it?
For large graphs (>Max.INT vertices) you want to have unique ids. 
Can you elaborate?

artem-aliev: Right, this limitation is still exists. It is GraphX one. But I do not expect any of graph algorithms will finish in reasonable time with >Max.INT vertices.

amuraru: Well - disagree, we have regular graphs with 10s of Billions of vertices.
Adding this limitation of 2,147,483,647 vertices will just limit the graphframes usage

SemyonSinchenko:
It looks for me completely outdated because PageRank does not require anymore sources to have only Int IDs. Changing partitioning of this functionality may break connected components and other places. I'm closing it now.

PR #245, opened 2017-11-29
Allow custom persistence level for internal graphframes dataframes

amuraru:
On large dataframes persisting on DISK only renders better performance in our case.
This patch adds support to set custom StorageLevel for internal DataFrames


SemyonSinchenko:
Closed in favor of #698

PR #259, opened 2018-01-22
Travis CI: download spark from apache mirror

darthsuogles:
Changed Travis dependency download script to use apache mirror CGI for Spark.
Also updated Spark versions used in the tests to match the latest patch release.

### DIFF
--- a/bin/download_travis_dependencies.sh
+++ b/bin/download_travis_dependencies.sh
@@ -1,34 +1,34 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -eux -o pipefail
 
 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
 
-function maybe_download_from_apache_mirror {
+function try_download_from_apache {
     local spark_tarball="${SPARK_BUILD}.tgz"
-    local apache_mirror_cgi="https://www.apache.org/dyn/closer.lua"
+    local apache_archive_prefix="https://archive.apache.org/dist"
     local spark_rel_path="spark/spark-${SPARK_VERSION}/${spark_tarball}"
-    local spark_url="${apache_mirror_cgi}?path=${spark_rel_path}"
+    local spark_bin_url="${apache_archive_prefix}/${spark_rel_path}"
+    local spark_md5_url="${apache_archive_prefix}/${spark_rel_path}.md5"
 
     echo "Downloading Spark if necessary"
     echo "Spark version = $SPARK_VERSION"
     echo "Spark build = $SPARK_BUILD"
-    echo "Spark build URL = $spark_url"
+    echo "Spark build URL = ${spark_bin_url}"
 
-    if [[ ! -f "${spark_tarball}" ]]; then
-        python -- << __PY_SCRIPT_EOF__ | xargs curl -O
-import sys, json
-json_str = """$(curl --silent --location "${spark_url}&asjson=1")"""
-pkg_info = json.loads(json_str)
-print("{}/${spark_rel_path}".format(pkg_info["preferred"]))
-__PY_SCRIPT_EOF__
+    # Existing files might be corrupt, clean them up.
+    rm -f "${spark_tarball}" "${spark_tarball}.md5"
 
-        echo "Content of directory:"
-        ls -la
-        tar xvf "${spark_tarball}" > /dev/null
-    fi
+    curl --retry 3 --retry-delay 7 -O "${spark_bin_url}"
+    curl --retry 3 --retry-delay 7 -O "${spark_md5_url}"
+
+    echo "Content of directory:"
+    ls -la
+    gpg --print-md MD5 "${spark_tarball}" | tee "${spark_tarball}.gen.md5"
+    diff "${spark_tarball}.gen.md5" "${spark_tarball}.md5"
+    tar -zxf "${spark_tarball}"
 }
 
 mkdir -p "${HOME}/.cache/spark-versions" && pushd $_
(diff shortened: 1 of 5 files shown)


bin/download_travis_dependencies.sh
```diff
@@ -1,18 +1,34 @@
 #!/usr/bin/env bash
 
-SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-${SPARK_VERSION}-bin-hadoop2.7.tgz"
+set -euo pipefail
```
felixcheung: consider: -x?

bin/download_travis_dependencies.sh
```diff
@@ -1,18 +1,34 @@
...
+        python -- << __PY_SCRIPT_EOF__ | xargs curl -O
+import sys, json
+json_str = """$(curl --silent --location "${spark_url}&asjson=1")"""
+pkg_info = json.loads(json_str)
+print("{}/${spark_rel_path}".format(pkg_info["preferred"]))
+__PY_SCRIPT_EOF__
+
+        echo "Content of directory:"
+        ls -la
+        tar xvf "${spark_tarball}" > /dev/null
```
felixcheung: might want to retry if/when curl or tar fails
it actually happens quite a bit

darthsuogles: That sounds great. I added retry and checksum, etc.

bin/download_travis_dependencies.sh
```diff
@@ -1,34 +1,34 @@
...
-set -euo pipefail
+set -eux -o pipefail
 
 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
 
-function maybe_download_from_apache_mirror {
+function try_download_from_apache {
     local spark_tarball="${SPARK_BUILD}.tgz"
-    local apache_mirror_cgi="https://www.apache.org/dyn/closer.lua"
+    local apache_archive_prefix="https://archive.apache.org/dist"
```
darthsuogles: Seems that it's easier to directly download from apache archive than from the mirror.

bin/download_travis_dependencies.sh
```diff
@@ -1,34 +1,34 @@
...
+
+    # Existing files might be corrupt, clean them up.
+    rm -f "${spark_tarball}" "${spark_tarball}.md5"
+
+    curl --retry 3 --retry-delay 7 -O "${spark_bin_url}"
+    curl --retry 3 --retry-delay 7 -O "${spark_md5_url}"
+
+    echo "Content of directory:"
+    ls -la
+    gpg --print-md MD5 "${spark_tarball}" | tee "${spark_tarball}.gen.md5"
```
darthsuogles: Spark distribution's `md5` is not generated from `md5sum`.

felixcheung:
That maybe, but there are some policy against direct download from Apache.org...

darthsuogles:
Ah, sorry, didn't realize the policy against direct download. 
Switching back to using the mirror and added MD5 files manually.

darthsuogles:
Strangely, the Spark version `2.1.1` I am downloading from the mirror gives a different checksum than the one from the archive.
```
< spark-2.1.1-bin-hadoop2.7.tgz: 52 A0 05 CE CB E0 CE 49  39 F5 5B 0C 78 65 3A 2C
---
> spark-2.1.1-bin-hadoop2.7.tgz: C0 AC C4 44 7D 5C D7 CB  43 5E 51 64 80 98 86 E1
```

felixcheung:
hmm, it should be, unless the mirror is out of sync or corrupted. does the tgz unpack?

felixcheung:
btw, I think it's a problem with 2.1.1 - it's gone from mirrors because the latest is 2.1.2
so your hash is probably a html page or something. eg. http://apache.claz.org/spark/

btw, we shouldn't need to keep our hashes. This is the way ASF release work
- official signed release is in https://dist.apache.org/repos/dist/release/spark/spark-2.1.2/spark-2.1.2-bin-hadoop2.7.tgz
- anything in the release share is replicated to mirrors, eg. http://apache.claz.org/spark/spark-2.1.2/spark-2.1.2-bin-hadoop2.7.tgz
- however, signature https://dist.apache.org/repos/dist/release/spark/spark-2.1.2/spark-2.1.2-bin-hadoop2.7.tgz.asc and hash https://dist.apache.org/repos/dist/release/spark/spark-2.1.2/spark-2.1.2-bin-hadoop2.7.tgz.sha should be checked against the release share (and the signing key https://dist.apache.org/repos/dist/release/spark/KEYS)

felixcheung:
for example https://github.com/apache/spark/blob/master/R/pkg/R/install.R#L214
but we don't check the hash there :) (maybe we should...)

darthsuogles:
Indeed. The file downloaded from the mirror wasn't `tgz`. 
Let me remove the signature check and perhaps put it in a later PR.

Thanks for pointing out the Apache release policies. 
If we were to check the

bin/download_travis_dependencies.sh
```diff
@@ -1,18 +1,40 @@
...
+__PY_SCRIPT_EOF__
+) | xargs curl --retry 3 --retry-delay 7 -O
+
+    echo "Content of directory:"
+    ls -la
+    tar -zxf "${spark_tarball}"
+}
+
+mkdir -p "${HOME}/.cache/spark-versions" && pushd $_
+try_download_from_apache || try_download_from_apache || try_download_from_apache
```
felixcheung: haha, neat with pipefail

PR #260, opened 2018-01-25
Try remove dangling Tokenizer call

darthsuogles:
I have no idea why we have a dangling `self._sc._jvm.org.apache.spark.ml.feature.Tokenizer()` that does not get assigned to anything. Are we including it to see if Spark ML can be successfully initialized?


felixcheung:
maybe? it was added here https://github.com/darthsuogles/graphframes/commit/d59f9578eb7d486ffadc6068f0b12c96b132e481
@jkbradley ?

felixcheung:
perhaps this is good to go?

PR #271, opened 2018-05-07
Fixes for Scala examples

alexott:
Changes include:
- add necessary imports
- remove not necessary type declarations
- make code `spark-shell` friendly
- fix other errors

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -115,7 +115,7 @@ as `vertices` and `edges` fields in the GraphFrame.
 
 <div data-lang="scala"  markdown="1">
 {% highlight scala %}
-import org.graphframes.examples
+import org.graphframes._
 val g: GraphFrame = examples.Graphs.friends  // get example graph
 
 // Display the vertex and edge DataFrames
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -331,8 +331,8 @@ def sumFriends(cnt: Column, relationship: Column): Column = {
 }
 //  (b) Use sequence operation to apply method to sequence of elements in motif.
 //      In this case, the elements are the 3 edges.
-val condition = Seq("ab", "bc", "cd").
-  foldLeft(lit(0))((cnt, e) => sumFriends(cnt, col(e)("relationship")))
+val condition = { Seq("ab", "bc", "cd")
+  .foldLeft(lit(0))((cnt, e) => sumFriends(cnt, col(e)("relationship"))) }
 //  (c) Apply filter to DataFrame.
 val chainWith2Friends2 = chain4.where(condition >= 2)
 chainWith2Friends2.show()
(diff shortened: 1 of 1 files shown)


docs/user-guide.md
```diff
@@ -416,14 +417,14 @@ triplets by using more complex motifs.
...
+val g = examples.Graphs.friends  // get example graph
 
 // Select subgraph based on edges "e" of type "follow"
 // pointing from a younger user "a" to an older user "b".
-val paths = g.find("(a)-[e]->(b)")
-  .filter("e.relationship = 'follow'")
-  .filter("a.age < b.age")
+val paths = g.find("(a)-[e]->(b)").
+  filter("e.relationship = 'follow'").
+  filter("a.age < b.age")
```
felixcheung: Spark code generally have the `.` on the next line, for example
https://github.com/apache/spark/blob/master/examples/src/main/scala/org/apache/spark/examples/ml/DecisionTreeExample.scala

docs/user-guide.md
```diff
@@ -260,10 +261,10 @@ For API details, refer to the [API docs](api/scala/index.html#org.graphframes.Gr
 
 {% highlight scala %}
 import org.graphframes.examples
-val g: GraphFrame = examples.Graphs.friends  // get example graph
+val g = examples.Graphs.friends  // get example graph
 
 // Search for pairs of vertices with edges in both directions between them.
-val motifs: DataFrame = g.find("(a)-[e]->(b); (b)-[e2]->(a)")
```
felixcheung: I think it's kinda useful to see the data type in doc text in somewhat unintuitive cases, like this one

docs/user-guide.md
```diff
@@ -383,8 +384,8 @@ The following example shows how to select a subgraph based upon vertex and edge
 
 <div data-lang="scala"  markdown="1">
 {% highlight scala %}
-import org.graphframes.examples
-val g: GraphFrame = examples.Graphs.friends
+import org.graphframes._
```
felixcheung: generally spark follows the guideline of not import ._ unless there are > 5 classes/types

docs/user-guide.md
```diff
@@ -678,8 +678,9 @@ val results2 = g.pageRank.resetProbability(0.15).maxIter(10).run()
 // Run PageRank personalized for vertex "a"
 val results3 = g.pageRank.resetProbability(0.15).maxIter(10).sourceId("a").run()
 
-# Run PageRank personalized for vertex ["a", "b", "c", "d"] in parallel
-val results3 = g.pageRank.resetProbability(0.15).maxIter(10).sourceIds(Array("a", "b", "c", "d")).run()
+// Run PageRank personalized for vertex ["a", "b", "c", "d"] in parallel
+// Works only in Spark 2.1+
```
felixcheung: this is true - I think we are missing some checks/handling for this in `parallelPersonalizedPageRank`

docs/user-guide.md
alexott: Ok, then I'll put explicit imports into it.

docs/user-guide.md
alexott: Ok, I've removed it because with current import, it won't work. I can update the code to return types...

docs/user-guide.md
felixcheung: I see, yes, that could be helpful. is there another way for this though?
like
```
val paths = (g.find()
                     .filter())
```

or that is ugly?

alexott: No, that's ok - I just pushed new commit, but used the `{}` to mark these as blocks of code...

alexott:
Please squash commits if you do merge...

felixcheung:
thanks!

Issue #63, opened 2016-03-15
Subgraph selection helper method


jkbradley:
It would be nice to have a helper method for selecting subgraphs.  I'm imagining something using the stdlib API:
```
def selectVertexSubgraph(expr): GraphFrame
def selectEdgeSubgraph(expr): GraphFrame
def selectSubgraph(vertexExpr, edgeExpr): GraphFrame
```
where ```expr``` is a Column or String for filtering.

These methods should ensure consistency between the resulting vertex, edge DataFrames.  E.g., if a subset of vertices are selected, then any edges connected with a dropped vertex should be dropped.

When selecting subsets of edges, it might also be nice to have options for choosing what to do with the vertices: Drop any vertices not connected to a selected edge, or keep all vertices?

jessyZu:
+1

PR #273, opened 2018-06-07
Add filter api

bohanjason:
Add three helper methods for selecting the subgraph:

filterVertices(expr): GraphFrame, This method filters vertices, and removes edges connected with a dropped vertex
filterEdges(expr): GraphFrame, This method filters edges, and keeps all vertices
dropIsolatedVertices(): GraphFrame, This method drops the isolated vertices that are not contained in any edge.

Fix issue #63: Subgraph selection helper method


src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -335,6 +336,51 @@ class GraphFrame private(
...
+   */
+  def filterVertices(condExpr: String): GraphFrame = filterVertices(expr(condExpr))
+
+  /** 
+   * Filter the edges according to Column expression, keep all vertices.
+   * @group subgraph
+   */
+  def filterEdges(condExpr: Column): GraphFrame = {
+    val vv = vertices
+    val ee = edges.filter(condExpr)
```
felixcheung: if we are trying to maintain consistency when filtering vertices, when we filter edges, don't we want to make sure there is no "disconnected vertices"?

felixcheung: why separately have dropIsolatedVertices?

felixcheung: or instead it should be an optional param to choose which way to go?

bohanjason: We have thought about the consistency for vertices. Instead of passing the optional parameter, we provide another function dropIsolatedVertices(). So users may use it if they want to maintain consistency. We think it's more clear compared to passing the optional flag. Besides, there is no cross- function optimization if we put these two functions together into one single function. 
```
filterEdges(cond).dropIsolatedVertices()
```

jkbradley: I like how the current PR handles consistency.
@felixcheung With regards to select vs filter,
* Do you prefer one name to the other?  I like "filter" more actually since it is similar to Dataset.filter.
* The current API supports both Columns and String expressions.

jkbradley:
I recommend renaming the arguments to match Dataset.filter:

In Scala:
condition: Column
conditionExpr: String

In Python:
condition

src/main/scala/org/graphframes/GraphFrame.scala
felixcheung: filter sounds good

felixcheung: so I think we need to consider a few more cases?

`dropIsolatedVertices` is one, what about edges connecting to vertices not in `g.vertices`?

I think there's going to be a good set of "get graph consistent" functions or "check graph for consistency" functions

MrBago: Thanks for taking a look @felixcheung.

It's easy enough to add an API for `dropInvalidEdges()`, but what other "graph consistency" use cases did you have in mind? If there are a good number, maybe we should open a new issue for this because it seems like the "filter graph" and the "graph consistency" use cases seem separable to me.

mengxr:
@felixcheung I think we can address the consistency issues in a follow-up PR, just to keep this one minimal.

+1 on @jkbradley 's suggestion to match Dataset.filter naming.

@bohanjason Could you update the PR and resolve conflicts with master? Let's get this PR in:)

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -360,6 +361,51 @@ class GraphFrame private(
...
+  def filterEdges(conditionExpr: String): GraphFrame = filterEdges(expr(conditionExpr))
+
+  /**
+   * Drop isolated vertices, vertices not contained in any edges.
+   * @group subgraph
+   */
+  def dropIsolatedVertices(): GraphFrame = {
+    val ee = edges
+    val e1 = ee.withColumn(ID, explode(array(col(SRC), col(DST))))
+    val vv = vertices.join(e1, Seq(ID), "left_semi")
```
mengxr: It will add duplicates. We need to add distinct to e1. Please also add a test.

bohanjason: I think the result will not include the duplicates if the vertices do not include duplicates . So we do not need an extra distinct operation ?

mengxr: you are right!

bohanjason:
@mengxr Fixed according to your comments

mengxr:
LGTM. Merged into master. Thanks!

Issue #276, opened 2018-06-13
Remove negation from motif finding DSL


jkbradley:
This proposal comes from thinking about https://github.com/graphframes/graphframes/issues/254

I recommend we remove negation from our DSL because:
* Its semantics are very unclear and prone to cause user confusion.
* Our DSL is loosely based on Cypher QL, which does not support negation.
* Users can handle negation on their own by combining results from multiple queries.

This is a breaking API change, but since we are in the 0.x release series, we can introduce this in a new minor release.

I wrote up more details in [this doc](https://docs.google.com/document/d/1e7HUdsTe7jNBjxYt2WflR0Pb5l3I9StjFplUitph6AE/); please check it out and comment in this Github issue as needed.

felixcheung:
my worry about removing this is also people seem to be using `!` pattern quite a bit. In the issue # 254 there are at least 3-4 distinct users using this (and running into the issue)

jkbradley:
Oh whoops!  I looked but missed this reference.  I also tried it out in their sandbox, and it seemed like it did not work.  I'll explore it more to see how/if they make the semantics clear.

jkbradley:
OK testing it manually, I've found:
* Negated (NOT) edges are allowed in WHERE statements
* All named vertex and edge elements in negated edges must appear in non-negated edges in the original MATCH statement
* Negated edges can have anonymous vertex and edge elements.
* If a negated edge an 0 named elements, then 0 edges are returned.  This means that the WHERE clause is interpreted as an AND.  E.g., the query `MATCH (a)-[]->(b)<-[]-(c); WHERE NOT ()-[]->()-[]->(); RETURN a,b` could be interpreted as: find 3-vertex sequences a->b<-c without edges going b->a or b->c.  However, Neo4J returns 0 results even when there are such sequences in my graph.

My new proposal: Keep negation, but restrict it as follows:
* Each negated edge must have at least 1 named element.
* Each named element in a negated edge must appear in a non-negated edge.

What do you think?

jkbradley:
UPDATE: Per our offline discussion, I updated [the doc](https://docs.google.com/document/d/1e7HUdsTe7jNBjxYt2WflR0Pb5l3I9StjFplUitph6AE/edit#).  Here is the gist:

## Proposed rules for motifs

The API guarantees to find all examples of named elements matching each term in the motif.
* A key implication here is that named vertices which only appear in negated terms will trigger outer joins.

[...]

PR #279, opened 2018-06-28
Fix negation in motif finding: helper methods

jkbradley:
This picks the helper methods and tests out from https://github.com/graphframes/graphframes/pull/278 for fixing negation in motif finding in https://github.com/graphframes/graphframes/issues/276

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -912,4 +912,12 @@ object GraphFrame extends Serializable with Logging {
     _broadcastThreshold = value
     this
   }
+
+  /**
+   * Return the given DataFrame with columns in a given order.
+   * @param names This method assumes that names is exactly the set of columns in the DataFrame.
+   */
+  private[graphframes] def enforceColumnOrder(df: DataFrame, names: Seq[String]): DataFrame = {
+    if (names.isEmpty) df else df.select(names.head, names.tail : _*)
+  }
 }
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -912,12 +912,4 @@ object GraphFrame extends Serializable with Logging {
     _broadcastThreshold = value
     this
   }
-
-  /**
-   * Return the given DataFrame with columns in the order specified by `names`.
-   * @param names This method assumes that names is exactly the set of columns in the DataFrame.
-   */
-  private[graphframes] def enforceColumnOrder(df: DataFrame, names: Seq[String]): DataFrame = {
-    if (names.isEmpty) df else df.select(names.head, names.tail : _*)
-  }
 }
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -914,7 +914,7 @@ object GraphFrame extends Serializable with Logging {
   }
 
   /**
-   * Return the given DataFrame with columns in a given order.
+   * Return the given DataFrame with columns in the order specified by `names`.
    * @param names This method assumes that names is exactly the set of columns in the DataFrame.
    */
   private[graphframes] def enforceColumnOrder(df: DataFrame, names: Seq[String]): DataFrame = {
--- a/src/main/scala/org/graphframes/pattern/patterns.scala
+++ b/src/main/scala/org/graphframes/pattern/patterns.scala
@@ -17,6 +17,7 @@
 
 package org.graphframes.pattern
 
+import scala.collection.mutable
 import scala.util.parsing.combinator._
 
 import org.graphframes.InvalidParseException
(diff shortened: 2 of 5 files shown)


src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -48,13 +49,124 @@ private[graphframes] object PatternParser extends RegexParsers {
...
+        }
+        edgeNames += name
+        addVertex(src)
+        addVertex(dst)
+      case AnonymousEdge(src, dst) =>
+        addVertex(src)
+        addVertex(dst)
+    }
+
+    patterns.foreach {
```
jkbradley: This method looks like it could be combined with the addVertex, addEdge logic above, but it's awkward to since we need to handle Negation differently.

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -912,4 +912,12 @@ object GraphFrame extends Serializable with Logging {
     _broadcastThreshold = value
     this
   }
+
+  /**
+   * Return the given DataFrame with columns in the order specified by `names`.
+   * @param names This method assumes that names is exactly the set of columns in the DataFrame.
```
yogeshg: The semantics of the method seem slightly different, it seems to assume that columns is a subset of those in Dataframe

jkbradley:
I pushed a rebase to make tests run again after removing Spark 2.0 support

src/main/scala/org/graphframes/GraphFrame.scala
mengxr: The doc doesn't explain the behavior when `names` is empty and it is not clearly stated by the method name.

src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -48,13 +49,124 @@ private[graphframes] object PatternParser extends RegexParsers {
...
+              s"anonymous negated edges !()-[]->().  Users can check for 0 edges in the graph " +
+              s"using the edges DataFrame.")
+          case e @ AnonymousEdge(_, _) =>
+            addEdge(e)
+        }
+      case AnonymousEdge(AnonymousVertex, AnonymousVertex) =>
+        throw new InvalidParseException(s"Motif finding does not support completely " +
+          s"anonymous edges ()-[]->().  Users can check for the existence of edges in the " +
+          s"graph using the edges DataFrame.")
+      case e @ AnonymousEdge(_, _) =>
```
mengxr: Do we support '()-[e]->()` (querying all edges)? Not sure which branch handles this case. May be useful to add a test.

src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -48,13 +49,124 @@ private[graphframes] object PatternParser extends RegexParsers {
...
+        addEdge(e)
+      case AnonymousVertex => // ok
+      case v @ NamedVertex(_) =>
+        addVertex(v)
+    }
+  }
+
+  /**
+   * Split patterns into positive (non-negated) and negated terms, and find all named vertices
+   * in each.
```
mengxr: The doc says "find all named vertices in each." This is a little misleading because Java/Scala takes the very first sentence as the method summary, but it is incomplete here.

src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -48,13 +49,124 @@ private[graphframes] object PatternParser extends RegexParsers {
...
+  }
+
+  /**
+   * Return the set of named vertices (and optionally edges) appearing in the given patterns,
+   * in the order they first appear in the sequence of patterns.
+   * @param includeEdges If true, include named edges in the returned sequence.
+   */
+  private[graphframes]
+  def findNamedElementsInOrder(patterns: Seq[Pattern], includeEdges: Boolean): Seq[String] = {
+    val elementSet = mutable.HashSet.empty[String]
```
mengxr: You may use LinkedHashSet (https://www.scala-lang.org/api/current/scala/collection/mutable/LinkedHashSet.html) to simply the code. It visits elements in the order they were inserted.

src/main/scala/org/graphframes/GraphFrame.scala
jkbradley: Actually, I forgot to remove this; it used to be used in several places.

jkbradley: Actually I might eliminate this.  It used to be used in several places but now only 1 (in the other PR).  There's not much value in keeping it as its own method now.

src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -48,13 +49,124 @@ private[graphframes] object PatternParser extends RegexParsers {
...
     }
+    assertValidPatterns(result)
+    result
+  }
+
+  /**
+   * Checks all Patterns for validity:
+   *  - Disallow named edges within negated terms
+   *  - Disallow term "()-[]->()" and its negation
+   *  - Disallow name to be shared by a vertex and an edge
```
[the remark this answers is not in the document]
jkbradley: That's acceptable (self-loop).  I'll add a test for it.

mengxr:
LGTM

PR #280, opened 2018-06-29
Remove Spark 2.0 support

jkbradley:
I'd like to remove Spark 2.0 support to help simplify fixing negation for motif finding.  The fix requires use of an explicit crossJoin call, but crossJoin was only added in Spark 2.1.  This patch removes 2.0 support.


Makefile
```diff
@@ -1,9 +1,9 @@
-all: 2.2.0s2.10 2.0.2 2.1.1 2.2.0 2.3.0
+all: 2.2.0s2.10 2.1.1 2.2.0 2.3.0
```
felixcheung: switch to 2.1.2? 2.2.1? 2.3.1? probably another PR.

PR #291, opened 2018-08-11
Remove 2.1 support

jkbradley:
I can't see us publishing another version for 2.1 (mainly since it hits too many issues when users try motif finding), so this PR shortens the build by removing that support.

### DIFF
--- a/python/graphframes/tests.py
+++ b/python/graphframes/tests.py
@@ -333,8 +333,6 @@ class GraphFrameLibTest(GraphFrameTestCase):
         self._hasCols(pr, vcols=['id', 'pagerank'], ecols=['src', 'dst', 'weight'])
 
     def test_parallel_personalized_page_rank(self):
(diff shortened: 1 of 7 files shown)


python/graphframes/tests.py
```diff
@@ -333,8 +333,6 @@ def test_page_rank(self):
         self._hasCols(pr, vcols=['id', 'pagerank'], ecols=['src', 'dst', 'weight'])
 
     def test_parallel_personalized_page_rank(self):
-        if not GraphFrameTestUtils.spark_at_least_of_version("2.1"):
```
ludatabricks: Do we want to remove all dependency of older version in the tests?
In `test_connected_components_friends`, we also check the version for 2.0.

jkbradley:
Thanks for reviewing this!

Re: quick-start.md: We should update the quick-start guide later since we don't have a new version to use in the old one's place.

I updated per your other comments.

jkbradley:
Thanks @ludatabricks !  Merging with master

PR #293, opened 2018-08-21
Update Travis to use Scala 2.11.8

jkbradley:
The Travis version was not being set correctly, causing build failures for Spark 2.2 with Python 3 (weirdly).  This PR is meant to fix that.


jkbradley:
will fix tomorrow...

jkbradley:
I checked this with Xiangrui offline, so I'll merge it now.

PR #294, opened 2018-08-21
use isSnapshot for version

MrBago:
fix - we never set isSnapshot in the build script.


jkbradley:
LGTM pending tests

MrBago:
On second thought lets not do this right now, we can do it when we bump the version number.

PR #295, opened 2018-08-21
bump version to 0.7.0-SNAPSHOT

jkbradley:


build.sbt
```diff
@@ -22,8 +22,12 @@ name := "graphframes"
 
 spName := "graphframes/graphframes"
 
+organization := "org.graphframes"
```
jkbradley: Just a nit, but we should include this

MrBago:
thanks @jkbradley, lgtm.

Issue #261, opened 2018-01-28
sbt.ResolveException: unresolved dependency while building Graphframes


debasishdebs:
Hi all,
I'm new to Spark, Pyspark and all its concepts. As a requirement for out Project, we have defided to go forward with Graphframes in our use case. I started on a naked system and did the following steps:

pip install pyspark==2.1.2 [Install Pyspark & Hence Spark]
SPARK_HOME=/root/anaconda3/envs/Py3.6/lib/python3.6/site-packages/pyspark
wget https://github.com/graphframes/graphframes/archive/release-0.5.0.zip [Download Graphframes in /opt/Downloads/ directory]
cd /opt/Downloads/
unzip release-0.5.0.zip [Unzipping]
cd graphframes-release-0.5.0/
build/sbt assembly.

I get the following error on running build [Copied on paste.org at this [Link ](https://paste2.org/hAj9fA0h)for easy readability]

I thought this is because I've slf4j dependency missing. I added the same for version 1.7.7 to my pom.xml and did mvn clean package.

Now, I've a .pom file as mentioned in traceback at the file specified, but it doesn't seem to be able to read it.

Is there anythin I'm doing wrong?

Thanks.

felixcheung:
Is org.slf4j#slf4j-api;1.7.7 there after it is added? 
according to build.sbt it should be  "scala-logging-api" % "2.1.2"

SemyonSinchenko:
It looks outdated. I'm closing it now, feel free to re-open in case you face this issue with the latest versions of spark and graphframes. There were a lot of updates to build.sbt recently

Issue #240, opened 2017-10-14
java.lang.UnsupportedClassVersionError: com/typesafe/scalalogging/Logger : Unsupported major.minor version 52.0


RayTsui:
When I execute the spark-shell in cluster and I input
val graphframes = GraphFrame.fromEdges(edges)
it gives me as follows:
java.lang.UnsupportedClassVersionError: com/typesafe/scalalogging/Logger : Unsupported major.minor version 52.0
at java.lang.ClassLoader.defineClass1(Native Method)
at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
at java.security.AccessController.doPrivileged(Native Method)
at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
... 50 elided

felixcheung:
it was suggested this can result from an older version of java - which version are you running with?

RayTsui:
The java version on cluster is 
java version "1.7.0_60-ea"
Java(TM) SE Runtime Environment (build 1.7.0_60-ea-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.60-b09, mixed mode)

Is something wrong?

felixcheung:
could you check your classpath and scala version?
perhaps com/typesafe/scalalogging/Logger  is there and not working with your java version?

RayTsui:
I have requested my colleagues to check the configuration in the cluster, I will post the feedback if the solution is available.

SemyonSinchenko:
I'm closing it now because it is very old. Feel free to re-open if the issue is still present in the latest versions of graphframes and spark.

Issue #113, opened 2016-10-06
SLF4J error


dvreed77:
I am running Scala 2.10.4 with Spark 1.5.0-cdh5.5.2 and am getting the following error when running a GraphFrames job:

scala

```
> val g = GraphFrame(v, e)
error: bad symbolic reference. A signature in Logging.class refers to type LazyLogging
in package com.typesafe.scalalogging.slf4j which is not available.
It may be completely missing from the current classpath, or the version on
the classpath might be incompatible with the version used when compiling Logging.class.
```

I am starting my spark-shell with the following command:

`spark-shell --jars /data/spark-jars/scalalogging-slf4j_2.10-1.1.0.jar,/data/spark-jars/graphframes-0.2.0-spark1.5-s_2.10.jar`

I have tried different versions of scalalogging, but nothing seems to work.

Thanks for the help.

thunterdb:
@dvreed77 we use the following package: `"com.typesafe.scala-logging" %% "scala-logging-slf4j" % "2.1.2"`, which you can use with `spark-shell` 's `--package` option

hxu:
This has broken the Python side APIs as well.  Should this not be packaged with the main graphframes jar?

hxu:
To get this working, I had to include both `scala-logging-slf4j` and `scala-logging-api` at version 2.1.2.  I set this by adding them to the `--jars` option to `spark-submit`

RayTsui:
Hi guys, 

I repeated your instructions to overcome the posted errors using commands as follows:
"spark-shell --master yarn --jars graphframes-0.5.0-spark2.1-s_2.11.jar scala-logging-slf4j_2.10-2.1.2.jar scala-logging-api_2.10-2.1.2.jar"

And then I execute the example graphframe code
"import org.graphframes._
// Create a Vertex DataFrame with unique ID column "id"
val v = sqlContext.createDataFrame(List(
  ("a", "Alice", 34),
  ("b", "Bob", 36),
  ("c", "Charlie", 30)
)).toDF("id", "name", "age")
// Create an Edge DataFrame with "src" and "dst" columns
val e = sqlContext.createDataFrame(List(
  ("a", "b", "friend"),
  ("b", "c", "follow"),
  ("c", "b", "follow")
)).toDF("src", "dst", "relationship")
// Create a GraphFrame
import org.graphframes.GraphFrame
val g = GraphFrame(v, e)"

it still gave me the following error:

[...]

PR #297, opened 2018-08-23
Remove scalalogging-slf4j from dependencies

tartakynov:
In this PR I've removed scala-logging-slf4j in favor of bare slf4j-api. 

GraphFrames depends on scala-logging v.2.1.2 which is incompatible with later scala-logging 3.x. Thus if other dependencies introduce scala-logging 3.x you get `java.lang.IncompatibleClassChangeError`. So you have to use shading to overcome this issue.

Simply upgrading scala-logging to 3.x will break Scala 2.10 cross-compilation for GraphFrames because 3.x is not available for Scala 2.10. So maybe it's better to remove scala-logging dependency in favor of bare slf4j-api, just like in Spark (see [this commit](https://github.com/apache/spark/commit/4c477117bb1ffef463776c86f925d35036f96b7a) in Spark)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ sparkComponents ++= Seq("graphx", "sql", "mllib")
 // add any Spark Package dependencies using spDependencies.
 // e.g. spDependencies += "databricks/spark-avro:0.1"
 
-libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.21"
+libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.16"
 
 libraryDependencies += "org.scalatest" %% "scalatest" % defaultScalaTestVer % "test"
 
--- a/build.sbt
+++ b/build.sbt
@@ -43,14 +43,9 @@ sparkComponents ++= Seq("graphx", "sql", "mllib")
 // add any Spark Package dependencies using spDependencies.
 // e.g. spDependencies += "databricks/spark-avro:0.1"
 
-libraryDependencies += "org.scalatest" %% "scalatest" % defaultScalaTestVer % "test"
+libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.21"
 
-// These versions are ancient, but they cross-compile around scala 2.10 and 2.11.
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -43,14 +43,9 @@ sparkComponents ++= Seq("graphx", "sql", "mllib")
 // add any Spark Package dependencies using spDependencies.
 // e.g. spDependencies += "databricks/spark-avro:0.1"
 
-libraryDependencies += "org.scalatest" %% "scalatest" % defaultScalaTestVer % "test"
+libraryDependencies += "org.slf4j" % "slf4j-api" % "1.7.21"
```
felixcheung: is there a reason for .21?
spark is running with 1.7.16
https://github.com/apache/spark/blob/9b6baeb7b9f73e1a38581f481ea7232db712deb8/pom.xml#L119

tartakynov: @felixcheung thanks, I totally missed that part, which is checking which version Spark is running. I've just picked up one the version which I've always used

tartakynov:
Okay, it passed

felixcheung:
thanks!

tmoschou:
Great to see this change. Unfortunately shading scala-logging 2.x didn't work for us, I presume because 2.x hardcodes package paths in in macros / quasi-quotes. 

https://github.com/lightbend/scala-logging/blob/v2.1.2/scala-logging-slf4j/src/main/scala_2.11/com/typesafe/scalalogging/slf4j/LoggerMacro.scala#L45

It would be great to see a new release of graphframes with this change. Is there a snapshots repository?

tartakynov:
@tmoschou @felixcheung perhaps it can be cherry-picked to 0.6.x release branch?

felixcheung:
I don't think we would do a 0.6.1 release?

tartakynov:
Why not? To me it looks like a legit minor release. Otherwise we would have to wait for months till the next major. I found a few open issues related to this fix https://github.com/graphframes/graphframes/issues/261 https://github.com/graphframes/graphframes/issues/240 https://github.com/graphframes/graphframes/issues/113

PR #304, opened 2018-12-04
[ML-5815] Pregel support (Scala API)

WeichenXu123:
Add Pregel interface to GraphFrames.

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -58,7 +58,7 @@ class GraphFrame private(
   }
 
   /** Get the `Pregel` object for running pregel.*/
-  val pregel = new Pregel(this)
+  def pregel = new Pregel(this)
 
   /**
    * Persist the dataframe representation of vertices and edges of the graph with the default
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -57,7 +57,11 @@ class GraphFrame private(
     "GraphFrame(v:" + v + ", e:" + e + ")"
   }
 
-  /** Get the `Pregel` object for running pregel.*/
+  /**
+   *  Get the `Pregel` object for running pregel.
+   *
+   *  See [[org.graphframes.Pregel]] for more details.
+   */
   def pregel = new Pregel(this)
 
   /**
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -57,6 +57,9 @@ class GraphFrame private(
     "GraphFrame(v:" + v + ", e:" + e + ")"
   }
 
+  /** Get the `Pregel` object for running pregel.*/
+  val pregel = new Pregel(this)
+
   /**
    * Persist the dataframe representation of vertices and edges of the graph with the default
    * storage level.
--- a/src/main/scala/org/graphframes/Pregel.scala
+++ b/src/main/scala/org/graphframes/Pregel.scala
@@ -107,7 +107,8 @@ class Pregel(val graph: GraphFrame) {
    * @param msgExpr The message expression. It is a sql expression and it
    *                can reference all propertis in the triplet, in the way
    *                `Pregel.src("src_col_name)`, `Pregel.edge("edge_col_name)`,
-   *                `Pregel.dst("dst_col_name)`
+   *                `Pregel.dst("dst_col_name)`. If `msgExpr` is null, pregel
+   *                will not send message.
    */
   def sendMsgToSrc(msgExpr: Column): this.type = {
     sendMsgs += Tuple2(Pregel.src(ID), msgExpr)
@@ -123,7 +124,8 @@ class Pregel(val graph: GraphFrame) {
    * @param msgExpr The message expression. It is a sql expression and it
    *                can reference all propertis in the triplet, in the way
    *                `Pregel.src("src_col_name)`, `Pregel.edge("edge_col_name)`,
-   *                `Pregel.dst("dst_col_name)`
+   *                `Pregel.dst("dst_col_name)`. If `msgExpr` is null, pregel
+   *                will not send message.
    */
   def sendMsgToDst(msgExpr: Column): this.type = {
     sendMsgs += Tuple2(Pregel.dst(ID), msgExpr)
--- a/src/main/scala/org/graphframes/Pregel.scala
+++ b/src/main/scala/org/graphframes/Pregel.scala
@@ -1,0 +1,223 @@
+package org.graphframes
+
+import java.io.IOException
+import java.util.UUID
+
+import org.apache.spark.sql.{Column, DataFrame}
+import org.apache.spark.sql.functions.{array, col, explode, struct, when}
+
+import org.apache.hadoop.fs.Path
+
+
+/**
+  * This class implement pregel on GraphFrame.
+  *
+  * We can get the Pregel instance by `graphFrame.pregel`, or construct it via a `graph`
+  * argument. and call a series of methods, then call method `run` to start pregel.
+  * It will return a DataFrame which is the vertices dataframe generated in the last round.
+  *
+  * When pregel `run` start, first, it will initialize some columns in vertices dataframe,
+  * which defined by `initVertexColumn` (user can call it multiple times),
+  * and then start iteration.
+  *
+  * Once the iteration start, in each supersteps of pregel, include 3 phases:
+  * phase-1) generate the `triplets` dataframe, and generate the “msg” to send.
+  * The target vertex to send and the msg is set via `sendMsg` method.
+  * phase-2) Do msg aggregation. It will use the aggregation column which is set via
+  * `aggMsgs` method. Each vertex aggregates those messages which it receives.
+  * Now vertices dataframe owns a new column which value is the aggregated msgs
+  * (received by each vertex). Now update vertex property columns, the update expressions
+  * is set by `updateVertexColumn` method and return the new vertices dataframe.
+  *
+  * The pregel iteration will run `maxIter` time, which can be set via `setMaxIter` method.
+  *
+  * @param graph The graph which pregel will run on.
+  */
+class Pregel(val graph: GraphFrame) {
+
+  private val withVertexColumnList = collection.mutable.ListBuffer.empty[(String, Column, Column)]
+
+  private var maxIter: Int = 100
+  private var checkpointInterval = 2
+
+  private var sendMsgs = collection.mutable.ListBuffer.empty[(Column, Column)]
+  private var aggMsgsCol: Column = null
+
+  private val CHECKPOINT_NAME_PREFIX = "pregel"
+
+  /** Set max iteration number for the pregel running. Default value is 100 */
+  def setMaxIter(value: Int): this.type = {
+    maxIter = value
+    this
+  }
+
+  /** Set the period to do the checkpoint when running pregel. Default value is 2 */
+  def setCheckpointInterval(value: Int): this.type = {
+    checkpointInterval = value
+    this
+  }
+
+  /**
+    * Use this method to set those vertex columns which will be initialized before
+    * pregel rounds start, and these columns will be updated after vertex receive
+    * aggregated messages in each round.
+    *
+    * @param colName the column name of initialized column in vertex Dataframe
+    * @param initialExpr The column expression used to initialize the column.
+    * @param updateAfterAggMsgsExpr The column expression used to update the column.
+    *                               Note that this sql expression can reference all
+    *                               vertex columns and an extra message column
+    *                               `Pregel.msgCol`.
+    */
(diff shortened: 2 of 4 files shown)


src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,234 @@
...
+
+    vertices
+  }
+
+}
+
+
+object Pregel extends Serializable {
+
+  val MSG_COL_NAME = "_msg"
```
MrBago: I believe we talked about using "_PREGEL_MSG_" during the design.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,234 @@
...
+  private val CHECKPOINT_NAME_PREFIX = "pregel"
+
+  /** Set max iteration number for the pregel running. Default value is 100 */
+  def setMaxIter(value: Int): this.type = {
+    maxIter = value
+    this
+  }
+
+  /** Set the period to do the checkpoint when running pregel. Default value is 2 */
+  def setCheckpointInterval(value: Int): this.type = {
```
MrBago: If value is 0, we don't checkpoint. Lets include that in the doc string.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,234 @@
...
+    val shouldCheckpoint = checkpointInterval > 0
+
+    while (iteration <= maxIter) {
+      val beginTime = System.currentTimeMillis()
+
+      val tripletsDF = vertices.select(struct(col("*")).as("src"))
+        .join(edges.select(struct(col("*")).as("edge")), Pregel.src("id") === Pregel.edge("src"))
+        .join(vertices.select(struct(col("*")).as("dst")), Pregel.edge("dst") === Pregel.dst("id"))
+
+      assert(sendMsgs.length > 0)
```
MrBago: Does this need to be in the loop? Can we raise a better error message if this is false.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,234 @@
...
+    var vertexUpdateColDF = graph.vertices.select((col("id") :: initVertexCols): _*)
+    var vertices = graph.vertices.join(vertexUpdateColDF, "id")
+    val edges = graph.edges
+
+    var iteration = 1
+
+    val shouldCheckpoint = checkpointInterval > 0
+
+    while (iteration <= maxIter) {
+      val beginTime = System.currentTimeMillis()
```
MrBago: This is never used? Did you mean to log start end end time for each iteration?

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,234 @@
...
+   * @param colName the column name of initialized column in vertex Dataframe
+   * @param initialExpr The column expression used to initialize the column.
+   * @param updateAfterAggMsgsExpr The column expression used to update the column.
+   *                               Note that this sql expression can reference all
+   *                               vertex columns and an extra message column
+   *                               `Pregel.msgCol`.
+   */
+  def withVertexColumn(colName: String, initialExpr: Column,
+                       updateAfterAggMsgsExpr: Column): this.type = {
+    withVertexColumnList += Tuple3(colName, initialExpr, updateAfterAggMsgsExpr)
```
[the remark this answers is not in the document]
WeichenXu123: yes.
Scala compiler will turn `a += b`  into `a.+=(b)` and compile it.
So if use `(...)`, then it become 3 arguments of `+=`, like `a.+=(b1, b2, b3)`, but what we need is `a.+=(Tuple3(b1, b2, b3))`

src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -57,6 +57,9 @@ class GraphFrame private(
     "GraphFrame(v:" + v + ", e:" + e + ")"
   }
 
+  /** Get the `Pregel` object for running pregel.*/
+  val pregel = new Pregel(this)
```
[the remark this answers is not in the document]
WeichenXu123: What's the "doctoring" for api page ?

MrBago:
@WeichenXu123 Thanks for the changes. This LGTM overall, I still think we could add a few more tests though.

MrBago:
One more small note, there are a few places where the documentation could be made more clear, but I think we could do this as a follow up PR.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+  def run(): DataFrame = {
+    val sendMsgsColList = sendMsgs.toList.map { case (id, msg) =>
+      struct(id.as("id"), msg.as("msg"))
+    }
+
+    val initVertexCols = withVertexColumnList.toList.map { case (colName, initExpr, _) =>
+      initExpr.as(colName)
+    }
+    val updateVertexCols = withVertexColumnList.toList.map { case (colName, _, updateExpr) =>
+      when(Pregel.msg.isNotNull, updateExpr).otherwise(col(colName)).as(colName)
```
mengxr: Just tell users if a vertex doesn't receive messages, its msg col will be null. We shouldn't add logic that keep the col(colName) unchanged in such case.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    var vertexUpdateColDF = graph.vertices.select((col("id") :: initVertexCols): _*)
+    var vertices = graph.vertices.join(vertexUpdateColDF, "id")
```
mengxr: * We don't need this self join. Just expand vertices.
* Do not reuse `vertices` name here.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+    val edges = graph.edges
+
+    var iteration = 1
+
+    val shouldCheckpoint = checkpointInterval > 0
+
+    require(sendMsgs.length > 0, "We need to set at least one message expression for pregel running.")
```
mengxr: move precondition checks to the beginning

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+
+    while (iteration <= maxIter) {
+      val tripletsDF = vertices.select(struct(col("*")).as("src"))
```
mengxr: minor: If this is equivalent to `Graph(vertices. edges).triplets`, let's use it.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+        // do checkpoint
+        newVertexUpdateColDF = newVertexUpdateColDF.checkpoint()
+        // TODO: remove last checkpoint file.
+      }
+      newVertexUpdateColDF.cache()
+
+      vertexUpdateColDF.unpersist()
+      vertexUpdateColDF = newVertexUpdateColDF
+
+      vertices = graph.vertices.join(vertexUpdateColDF, "id")
```
mengxr: move `updateVertexCols` here

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+      var msgDF: DataFrame = tripletsDF
+        .select(explode(array(sendMsgsColList: _*)).as("msg"))
+        .select(col("msg.id"), col("msg.msg").as(Pregel.MSG_COL_NAME))
+
+      val newValueDF = msgDF
+        .filter(Pregel.msg.isNotNull)
+        .groupBy("id")
+        .agg(aggMsgsCol.as(Pregel.MSG_COL_NAME))
+
+      val verticesWithMsg = vertices.join(newValueDF, Seq("id"), "left_outer")
```
mengxr: do not join old vertices here

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * The pregel iteration will run `maxIter` time, which can be set via `setMaxIter` method.
+ *
+ * @param graph The graph which pregel will run on.
+ */
+class Pregel(val graph: GraphFrame) {
+
+  private val withVertexColumnList = collection.mutable.ListBuffer.empty[(String, Column, Column)]
+
+  private var maxIter: Int = 100
```
mengxr: make the default smaller, e.g., 10

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+   * (src-edge-dst) will generate zero or one message and the message will
+   * be sent to the src vertex of this triplet.
+   *
+   * @param msgExpr The message expression. It is a sql expression and it
+   *                can reference all propertis in the triplet, in the way
+   *                `Pregel.src("src_col_name)`, `Pregel.edge("edge_col_name)`,
+   *                `Pregel.dst("dst_col_name)`
+   */
+  def sendMsgToSrc(msgExpr: Column): this.type = {
+    sendMsgs += Tuple2(Pregel.src("id"), msgExpr)
```
mengxr: Avoid using literal strings in code. Define constants.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+   * @param colName the column name of initialized column in vertex Dataframe
+   * @param initialExpr The column expression used to initialize the column.
+   * @param updateAfterAggMsgsExpr The column expression used to update the column.
+   *                               Note that this sql expression can reference all
+   *                               vertex columns and an extra message column
+   *                               `Pregel.msgCol`.
+   */
+  def withVertexColumn(colName: String, initialExpr: Column,
+                       updateAfterAggMsgsExpr: Column): this.type = {
+    withVertexColumnList += Tuple3(colName, initialExpr, updateAfterAggMsgsExpr)
```
mengxr: Verify if `colName` exists.

src/main/scala/org/graphframes/Pregel.scala
```diff
@@ -0,0 +1,236 @@
...
+   * Set the message column. In each round of pregel, each triplet
+   * (src-edge-dst) will generate zero or one message and the message will
+   * be sent to the src vertex of this triplet.
+   *
+   * @param msgExpr The message expression. It is a sql expression and it
+   *                can reference all propertis in the triplet, in the way
+   *                `Pregel.src("src_col_name)`, `Pregel.edge("edge_col_name)`,
+   *                `Pregel.dst("dst_col_name)`
+   */
+  def sendMsgToSrc(msgExpr: Column): this.type = {
```
mengxr: Mention `null` value.

src/main/scala/org/graphframes/Pregel.scala
WeichenXu123: If this, I try this way:
```
aggMsgDF = ...
// do cache & checkpoint on `aggMsgDF`
val verticesWithMsg = vertices.join(aggMsgDF, Seq(ID), "left_outer")
vertices = verticesWithMsg.select((originalVertexCols ++ updateCols): _*)
```
But it bring issues that current iteration vertices DF depends on previous one and the dependency cannot be cut by the cache/persist.

So, 
 - if we only cache the aggregated msg DF in each iteration, then we can remove one join here, **but it bring issues that the DF dependency cannot be cut by the cache/checkpoint and cause computation lineage to be long.**
 - if we only want to cache the additional columns for vertex DF in each iteration, then we have to do the two joins (one before cache/persist, one after it)
 - if we cache the whole vertices (including original columns and additional columns), then we can remove one join here.

I try every way but I do not find the best solution which can satisfy:
 - only cache additional columns or only cache aggregated msg column
 - remove one join here
 - do not cause computation lineage increased in each iteration.

Could you help think of a better solution (with pseudo code) ?

WeichenXu123:
I address all comments, except:
 - move "updateExpr" to be after "checkpoint" and eliminate one join. This need more discussion. I can't figure out a good solution here.
 - the `withVertexColumn`'s `updateAfterAggMsgsExpr` null value, whether we regard null value as "1) keep old value and do not update" or "2) update null value as a new value" ? I think option 2) may be better because sometimes user may need to update columns to be null, and if user want to keep old value use can use `col(thisColName)` to reference old value.

@MrBago @mengxr What about merge this first and @MrBago can do some doc refining and I can add some possible improvement as following PRs ?

src/main/scala/org/graphframes/Pregel.scala
mengxr: The logic of not updating the "rank" column when there is no msg received is not consistent with Pregel algorithm definition. For example, in page rank, a node without any incoming edge will lose the rank after one iteration (or the reset probability). We cannot keep its rank unchanged in that case. Leave it to users.

mengxr:
I'm merging this in to unblock follow-up work. @WeichenXu123 Could you submit a follow-up PR to fix the logic of handling no messages? Thanks!

PR #306, opened 2018-12-12
[ML-5815][FOLLOW-UP] Update pregel testcase "pageRank"

WeichenXu123:
Update pregel testcase "pageRank". When vertex receive no message, then its rank will be zero.

### DIFF
--- a/src/test/scala/org/graphframes/PregelSuite.scala
+++ b/src/test/scala/org/graphframes/PregelSuite.scala
@@ -17,56 +17,44 @@
 
 package org.graphframes
 
+import org.scalactic.Tolerance._
+
 import org.apache.spark.sql.functions._
 
 class PregelSuite extends SparkFunSuite with GraphFrameTestSparkContext {
 
   import testImplicits._
 
-  def almostEqual(a: Double, b: Double, relTol: Double): Boolean = {
-    return math.abs(a - b) <= math.abs(a) * relTol
-  }
-
   test("page rank") {
+    val edges = Seq(
+      (0L, 1L),
+      (1L, 2L),
+      (2L, 4L),
+      (2L, 0L),
+      (3L, 4L), // 3 has no in-links
+      (4L, 0L),
+      (4L, 2L)
+    ).toDF("src", "dst").cache()
+    val vertices = GraphFrame.fromEdges(edges).outDegrees.cache()
+    val numVertices = vertices.count()
+    val graph = GraphFrame(vertices, edges)
 
-    val vecDF1 = Seq(
-      (0L, 1),
-      (1L, 1),
-      (2L, 2),
-      (3L, 0),
-      (4L, 3)
-    ).toDF("id", "outDegree")
-
-    val edgeDF1 = Seq(
-      (0L, 1L, 2.0),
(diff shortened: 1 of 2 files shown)


src/test/scala/org/graphframes/PregelSuite.scala
```diff
@@ -58,14 +58,14 @@ class PregelSuite extends SparkFunSuite with GraphFrameTestSparkContext {
     val pageRankResultDF = new Pregel(graph1)
       .setMaxIter(15)
       .withVertexColumn("rank", lit(1.0 / N),
-        when(Pregel.msg.isNotNull, Pregel.msg).otherwise(col("rank")))
+        when(Pregel.msg.isNotNull, Pregel.msg).otherwise(lit(0.0)))
```
mengxr: `coalesce(Pregel.msg, lit(0.0))`

WeichenXu123: I use `(1.0 - alpha) / N` instead of zero when recv no msgs.
because the page rank update formula is `new_pr = sum(msgs) * alpha + (1 - alpha) / N`

src/test/scala/org/graphframes/PregelSuite.scala
```diff
@@ -58,14 +58,14 @@ class PregelSuite extends SparkFunSuite with GraphFrameTestSparkContext {
     val pageRankResultDF = new Pregel(graph1)
       .setMaxIter(15)
       .withVertexColumn("rank", lit(1.0 / N),
-        when(Pregel.msg.isNotNull, Pregel.msg).otherwise(col("rank")))
+        coalesce(Pregel.msg, lit(1.0 - pr_alpha) / N))
```
mengxr: minor: `pr_alpha` -> `alpha` or `prAlpha`

src/test/scala/org/graphframes/PregelSuite.scala
```diff
@@ -58,14 +58,14 @@ class PregelSuite extends SparkFunSuite with GraphFrameTestSparkContext {
...
       .sendMsgToDst(col("src.rank") / col("src.outDegree"))
       .aggMsgs(sum(col(Pregel.MSG_COL_NAME)) * lit(pr_alpha) + lit(1.0 - pr_alpha) / N)
       .run()
 
     val res = pageRankResultDF.sort(col("id"))
       .select("rank").as[Double].collect()
-    res.zip(Array(0.194, 0.195, 0.252, 0.194, 0.2)).foreach { case (a, b) =>
+    res.zip(Array(0.095, 0.111, 0.133, 0.095, 0.030)).foreach { case (a, b) =>
```
mengxr: The new result seems wrong to me because the ranks should sum up to 1.0. Could you double check and add a test on the sum?

mengxr:
The main issue was previous test graph contains a vertex with no out-links. Ideally, its rank should be distributed to all vertices in each iteration (reset). Another workaround is to send it back via in-links. We are testing the Pregel impl instead of implementing Page Rank. So I simplified the graph, only left one node with no in-links, which can be handled by Pregel easily.

PR #307, opened 2018-12-12
[ML-5886] Allow building docs in docker

MrBago:
Building docs for graphframes requires multiple dependences. Moving the build to docker allows us to more quickly and consistency  build our docs for release.


Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:16.04
+
+ARG PYTHON_VERSION=3.6
+
+RUN apt-get update && \
+    apt-get install -y wget bzip2 build-essential openjdk-8-jdk ssh sudo && \
+    apt-get clean
+
+# Add ubuntu user and enable password-less sudo
```
mengxr: We don't need to add ubuntu user to build the docs.

dev/build-docs-in-docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+cd ./docs
+SKIP_SCALADOC=1 PRODUCTION=1 jekyll build
```
mengxr: We need to generate Scala API docs.

dev/build-docs.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+# build the base image
+docker build -t databricks/graphframes --build-arg PYTHON_VERSION=2.7 .
```
mengxr: Do we need Python 2.7?

docs/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# An docker image `databricks/sparkdl` need to be generated by run
+#     $docker build -t databricks/sparkdl --build-arg PYTHON_VERSION=2.7 .
+# on base directory. Currently we need to build it with Python2.
+
+FROM databricks/graphframes
+
+RUN sudo apt-get install -y python-pip && \
```
mengxr: * `sudo` is not needed
* Always do

~~~
apt-get updates &&
...
apt-get clean
~~~

Because docker will cache the layer and the apt cache might not be not up-to-date.

docs/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    sudo pip install --upgrade pip && \
+    sudo pip install sphinx
```
mengxr: maybe we want to fix the version

docs/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN apt-get update && \
```
mengxr: Put all `apt-get` in a single `RUN` so it only creates a single layer.

docs/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    apt-get install -y ruby ruby-dev && \
+    apt-get clean
+
+RUN gem install jekyll bundler
+RUN gem install jekyll-redirect-from pygments.rb
```
mengxr: minor: maybe merge the two `RUN`s

dev/build-docs.sh
MrBago: I used python2.7 because this is in the sparkdl docs/Dockerfile

# An docker image `databricks/sparkdl` need to be generated by run
#     $docker build -t databricks/sparkdl --build-arg PYTHON_VERSION=2.7 .
# on base directory. Currently we need to build it with Python2.

mengxr: Yes, because we cannot build with Python 3 there. If we have to use Python 2 here, it would be nice to add an inline comment to explain why we don't use the default one.

dev/build-docs.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# We use python2.7 for building the docs because Pygments.rb/Pygments is incompatible with
+# python 3, more info: https://github.com/jekyll/jekyll/issues/2604
+docker build -t databricks/graphframes --build-arg PYTHON_VERSION=2.7 .
+
+# build the docs image
+docker build -t databricks/graphframes-docs docs/
+
+# create the assembly jar on host because we have ivy/maven cache
+build/sbt 'set test in assembly := {}' assembly
```
mengxr: The `jekyll build` command will trigger `build/sbt clean compile doc` in `_plugins/copy_api_dirs.rb`. The assembly built here will get cleaned. Let's do the following:

* Remove this line.
* Mount "\~/.ivy2" and "\~/.sbt" to accelerate the build inside docker:

~~~
docker run --rm -v "$(pwd):/mnt/graphframes" -v "$HOME/.sbt:/root/.sbt" -v "$HOME/.ivy2:/root/.ivy2" ...
~~~

mengxr:
LGTM pending tests

mengxr:
Merged since the tests do not touch the doc build.

PR #308, opened 2018-12-13
[ML-5893] Support Spark 2.4, Python 3.6, and remove Spark 2.2

mengxr:
* build with Spark 2.4 by default, remove Spark 2.2 support
* test against Python 3.6 instead of 3.5
* remove helpers (no longer needed with 2.3/2.4)
* re-enable parallel page rank test w/ Spark 2.4, added a test library to compare versions
* remove Makefile (not used/documented)

cc: @WeichenXu123 @felixcheung


mycaule:
Hello,

Do you support Scala 2.12 (since Spark 2.4 does) ?

Thank you.

PR #313, opened 2019-01-05
[ML-5815] [FOLLOW-UP] Change pregel default checkpointInterval to be 1

WeichenXu123:
Change pregel default checkpointInterval to be 1.

In my test, this will reduce SQL planner time in several cases.
And localCheckpoint is very fast so more frequent checkpoint won't increase execution time too much.


felixcheung:
this seems aggressive? ...

mengxr:
@WeichenXu123 Could you post the numbers from performance tests? You should also update the default value in Python.

@felixcheung The plan optimization time is indeed significant.

PR #319, opened 2019-01-16
[ML-5967] Publish with spDist

MrBago:
Use spDist with manual upload to publish instead of spPublish because the upload step of spDist if flaky.


yogeshg:
LGTM

PR #344, opened 2019-10-09
Building against spark-3.0.0-snapshot

WeichenXu123:
Building against spark-3.0.0-snapshot.

Fix some issue under spark-3.0.0-snapshot:
* removed method `SQLContext.getOrCreate`
* removed column operator `!==`
* some other unsupported syntax in scala 2.12

Fix a problematic python test `test_label_progagation`

Update travis to test against spark 2.4.4 and spark 3.0.0 snapshot.

WIP: (skip python test for now because I don't find spark nightly build tarball downloading address)


WeichenXu123:
Note: My local test is OK for spark snapshot version.
The travis failed because sbt has some issue when dealing with "SNAPSHOT" dependencies.

.travis.yml
```diff
@@ -10,25 +10,15 @@ cache:
...
   include:
-    - env: PYSPARK_PYTHON=python2 SPARK_VERSION=2.3.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python2 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python3 SPARK_VERSION=2.3.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python3 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-
-before_install:
- - ./bin/download_travis_dependencies.sh
+    - env: PYSPARK_PYTHON=python2 SCALA_VERSION=2.11.8 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
+    - env: PYSPARK_PYTHON=python2 SCALA_VERSION=2.12.10 SPARK_VERSION=3.0.0-SNAPSHOT SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
```
jkbradley: I'd recommend we drop this since Spark 3.0 is expected to deprecate python 2 support.

.travis.yml
```diff
@@ -10,25 +10,15 @@ cache:
...
 matrix:
   include:
-    - env: PYSPARK_PYTHON=python2 SPARK_VERSION=2.3.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python2 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python3 SPARK_VERSION=2.3.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - env: PYSPARK_PYTHON=python3 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-
-before_install:
- - ./bin/download_travis_dependencies.sh
+    - env: PYSPARK_PYTHON=python2 SCALA_VERSION=2.11.8 SPARK_VERSION=2.4.4 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
```
jkbradley: I think it's OK if we don't bother building against 2.4 as well.  The current version should continue to work, and if it breaks, we can always create a 2.4-compatible branch off of an old commit.

WeichenXu123:
Note: 
I introduce a repo "https://repository.apache.org/snapshots/" which is for spark SNAPSHOT dependency. Now sbt building works fine.

Note 2:
Graphframes python test are skip now, because I don't find the downloading address for Spark 3.0.0-SNAPSHOT distribution tarball (which graphframes python test require it) @jkbradley @mengxr Who should I ask to address this issue ?

.travis.yml
jkbradley: Fair enough; I'm OK with keeping it for now, but if we have to do a lot of work to maintain 2.4 compatibility, we'd probably end up dropping 2.4 support.

jkbradley:
This LGTM other than disabling python tests.  I recommend waiting to merge this until we can re-enable those.  I'll message about that offline.

mengxr:
@WeichenXu123 Please build a snapshot distribution from Spark master and upload it to a public S3 bucket to be used in CI. We need to improve the process but it is good enough for now.

python/graphframes/examples/belief_propagation.py
```diff
@@ -19,7 +19,6 @@
 
 from pyspark import SparkConf, SparkContext
 from pyspark.sql import SQLContext, functions as sqlfunctions, types
-from pyspark.tests import QuietTest as SuppressSparkLogs
```
WeichenXu123: The class `QuietTest` is annoying. In spark 2.4 it is in `pyspark.tests` package but in spark 3.0 it is in pyspark.testing` package.
So I remove it from this file. The example code here should not use some class inside pyspark test.

WeichenXu123:
@jkbradley Ready to merge.

mengxr:
LGTM. Merged into master.

PR #347, opened 2019-10-14
Remove unnecessary count and log statements

dennislloydjr:
Removes a count and log statement that have negative performance impacts and result only in some logging.

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -308,8 +308,6 @@ object ConnectedComponents extends Logging {
     val g = prepare(graph)
     val vv = g.vertices
     var ee = g.edges.persist(intermediateStorageLevel) // src < dst
-    val numEdges = ee.count()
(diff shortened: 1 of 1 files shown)


RyanBeatty:
Is there any way to help move this PR along and get it merged? I'm running some large graph computations at work and noticed that this `count()` statement was taking up a significant amount of time (at great cost to us)

sumersao:
+1 On helping to move it along

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -308,8 +308,6 @@ object ConnectedComponents extends Logging {
     val g = prepare(graph)
     val vv = g.vertices
     var ee = g.edges.persist(intermediateStorageLevel) // src < dst
-    val numEdges = ee.count()
```
SemyonSinchenko, line 311: This count is needed to materialize previously persisted edges DataFrame. Why is it considered as unnecessary? cc: @SauronShepherd

SemyonSinchenko:
Closed. The reason is in the last two comments.

PR #355, opened 2020-02-03
removed expensive computation from connected components

kundan59:
These lines in Connected Components are only used for logging and the computation of the count can be rather expensive:

    val numEdges = ee.count()
    logInfo(s"$logPrefix Found $numEdges edges after preparation.")


mengxr:
count is to persist the edges. if we cannot persist all edges, the computation would be much slower.

PR #363, opened 2020-04-20
Remove the use of deprecated methods and warnings

mengxr:
* Untyped Scala UDF is disabled at runtime by default in Spark 3.0. I changed the UDF creation in ShortestPath to use Java UDF with SQL types to bypass it.
* Update patterns.scala to avoid a compile time warning.
* Update build.sbt to show all deprecation/feature warnings.
* Update log4j to hide warnings from CacheManager during tests.


kujon:
Any indication of when this is going to get released?

PR #370, opened 2020-08-17
Quote column names with backticks

EnricoMi:
This guards against columns that contain dots. These column names must be quoted to avoid that Spark parses them again.

Fixes following exception when your vertices DataFrame contains columns with dots (e.g. `"a.name"`):

```
Cannot resolve column name "a.name" among (id, a.name); did you mean to quote the `a.name` column?;
org.apache.spark.sql.AnalysisException: Cannot resolve column name "a.name" among (id, a.name); did you mean to quote the `a.name` column?;
	at org.apache.spark.sql.Dataset.$anonfun$resolve$1(Dataset.scala:270)
	at scala.Option.getOrElse(Option.scala:189)
	at org.apache.spark.sql.Dataset.resolve(Dataset.scala:263)
	at org.apache.spark.sql.Dataset.col(Dataset.scala:1353)
	at org.apache.spark.sql.Dataset.apply(Dataset.scala:1320)
	at org.graphframes.GraphFrame$.$anonfun$nestAsCol$1(GraphFrame.scala:820)
	...
```

Fixes this issue also in other places.

### DIFF
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -882,7 +882,8 @@ object GraphFrame extends Serializable with Logging {
    * Note: This can be replaced with org.apache.spark.sql.catalyst.util.QuotingUtils.quoteIfNeeded
    * once support for Spark 3 has been dropped
    */
-  private[graphframes] def quote(column: String): String = s"`$column`"
+  private[graphframes] def quote(column: String): String =
+    s"`${column.replace("`", "``")}`"
 
   /**
    * Helper for column names containing a dot. Quotes the given column name with backticks to
--- a/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/src/main/scala/org/graphframes/GraphFrame.scala
@@ -55,9 +55,11 @@ class GraphFrame private (
   override def toString: String = {
     // We call select on the vertices and edges to ensure that ID, SRC, DST always come first
     // in the printed schema.
-    val vCols = (ID +: vertices.columns.filter(_ != ID).toIndexedSeq).map(col)
+    val vCols = (ID +: vertices.columns.filter(_ != ID).toIndexedSeq).map(quote).map(c => col(c))
     val eCols =
-      (SRC +: DST +: edges.columns.filter(c => c != SRC && c != DST).toIndexedSeq).map(col)
+      (SRC +: DST +: edges.columns.filter(c => c != SRC && c != DST).toIndexedSeq)
+        .map(quote)
+        .map(c => col(c))
     val v = vertices.select(vCols.toSeq: _*).toString
     val e = edges.select(eCols.toSeq: _*).toString
     "GraphFrame(v:" + v + ", e:" + e + ")"
@@ -356,7 +358,7 @@ class GraphFrame private (
     val df = findSimple(augmentedPatterns)
 
     val names = Pattern.findNamedElementsInOrder(patterns, includeEdges = true)
-    if (names.isEmpty) df else df.select(names.head, names.tail: _*)
+    if (names.isEmpty) df else df.select(quote(names.head), names.tail.map(quote): _*)
   }
 
   // ======================== Other queries ===================================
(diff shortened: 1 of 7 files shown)


EnricoMi:
@mengxr @felixcheung how do you feel about this fix?

EnricoMi:
@mengxr @felixcheung @WeichenXu123 ping

rjurney:
Oh thank you, I hate transforming to underscores!

EnricoMi:
Rebased. Thanks for looking into this.

SemyonSinchenko (approved):

SemyonSinchenko:
@rjurney @bjornjorgensen Hello! What do you think about this one? For me it looks very cool, I think it is very nice to have feature for GraphFrames!

bjornjorgensen:
.. and it is a breaking change to.

SemyonSinchenko:
> .. and it is a breaking change to.

To be honest I don't see why is it a breaking change? For end users everything should be exactly the same, until they are doing something similar on their side. From the other side complex pipelines with graphframes with column-names generation and similar things it should be much more smooth with that change.

bjornjorgensen:
> > .. and it is a breaking change to.
> 
> To be honest I don't see why is it a breaking change? For end users everything should be exactly the same, until they are doing something similar on their side. From the other side complex pipelines with graphframes with column-names generation and similar things it should be much more smooth with that change.

What you don't see any braking changes, but you explain them in the 3 next lines! 

I don't understand what this have to do with graphframes as this project code..

bjornjorgensen:
> > .. and it is a breaking change to.
> 
> You mean if column names are backticked already? That should be doable.
> 
yes, and I think that it's very common to have.  


> Any other use case that breaks due to this change?

Not that I can think of right now.

bjornjorgensen:
like pandas json_normalize uses sep='.' as default. https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html

EnricoMi:
> but its something that should have been in spark not here..
> .. and it is a breaking change to.

Looking more closely, this is not a breaking change. And Spark API expects the calling site to properly escape column names if they contain dots (unless they reference nested columns). Spark won't change that, otherwise, that would be a breaking change on their side. Since Spark 4.0.0, they even provide methods in `QuotingUtils` that can be used in user code (similar to `quote` implemented here).

Backticks are only required when referencing a column by its name (`select`, `col`, ...). A column name in the schema is not escaped by backticks. This pull requests only touches code where Graphframes references columns by their name, and that name is taken from the schema (not the user calling Graphframes API). Hence, the user cannot provide backticed column names, Graphframes is responsible for properly escaping the column names.

Without this fix, users of Graphframes have to workaround this by renaming affected columns:
https://github.com/G-Research/spark-dgraph-connector/blob/09dacbf/src/main/scala/uk/co/gresearch/spark/dgraph/graphframes/package.scala#L42-L44

SemyonSinchenko:
@bjornjorgensen Did you have a chance to check the latest commits? It seems to me that it should cover all of your concerns and nothing else is blocking merging. Thanks in advance!

bjornjorgensen:
hm... well I think its ok.. 

If we did have branches for graphframes we didn't needed this for branch 4 

I think we can have it, but if there are some user problems with it we will need to remove it.

SemyonSinchenko:
@rjurney What do you think about it? For me it is ready to merge.

EnricoMi:
> If we did have branches for graphframes we didn't needed this for branch 4

With "branch 4" I presume you refer to Spark4. What is different there? Do you imply backticks are not needed with Spark4? How are ambiguous column names resolved there?

EnricoMi:
> Since Spark 4.0.0, they even provide methods in [QuotingUtils](https://github.com/apache/spark/blob/523ff15/sql/api/src/main/scala/org/apache/spark/sql/catalyst/util/QuotingUtils.scala#L23) that can be used in user code

With "user code" I meant code that uses the Spark API, which in our case is Graphframes code. So once Spark3 support is dropped by Graphframes, the `quote` implementation of this PR can be removed and replaced with calling into `QuotingUtils`. So most of this PR remains.

SemyonSinchenko:
@bjornjorgensen Were your concerns resolved by the latest commits? If so, could you approve the PR?

cc: @rjurney @SauronShepherd could you take a look too? If you are fine with it, I would like to resolve conflicts and merge it. I see it as a very important feature.

EnricoMi:
Conflict resolved.

SemyonSinchenko (approved):

SemyonSinchenko:
> Shouldn't we add some tests to check that GraphFrames actually works fine with nested columns?

There are two tests in this PR already, do you think there should be more?

SemyonSinchenko:
@SauronShepherd Hi! Were your concerns resolved? I would like to merge this one to avoid even more merge conflicts, because this PR changes a lot of files.

SauronShepherd:
Sorry, I didn't have time to look at it. Go ahead.

SemyonSinchenko:
Thanks for the contribution @EnricoMi !

PR #377, opened 2020-11-30
Correct the order of the analogy

johnnyleitrim:
When making an analogy (A is to a as B is to b), the first part (A) is the part you _don't know_ and the second part (B) is the part you know.

This sentence is trying to explain GraphFrames to people who already know GraphX, so GraphX is last in the analogy.


mengxr (approved):
Thanks for the correction!

PR #381, opened 2021-03-09
Extend to Spark 3.1

EnricoMi:
GraphFrames work fine with Spark 3.1, so I think it should also support the latest Spark version.


mengxr:
LGTM. Thanks!

rjurney:
@mengxr now we need a release cut, right?

Issue #384, opened 2021-04-08
Since bintray repo is shutting down. Not able to download this jar.


skb81230:
HI,

SInce bintray repo is down, i tried to add ICM repo. But still not able to downlad this jar. Can someone please help.

Could not resolve dependencies for project com.reltio:analytics:jar:2021.1-develop-SNAPSHOT: Could not find artifact graphframes:graphframes:jar:0.5.0.3-spark2.2 in icm (http://maven.icm.edu.pl/artifactory/repo/)

mheilman:
bintray is going away on May 1, 2021, according to [this page](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).

skb81230:
Hi Michael,

Thanks for replying. Yes Bintray repo is shutting down.
Please help where we can find the graph frames jars ? Any other Repo?

mheilman:
Sorry, I don't know where the package will be available after bintray goes down.

skb81230:
Thank you for your response

nvander1:
@jkbradley @thunterdb Any update on this? The sunset is coming soon, and the other brownout is today.

DCameronMauch:
Yep, I am having the same issue.  Can't find a JAR for this online anywhere.

nvander1:
What is the plan for this? Builds are failing now because the dependency can no longer be found

unresolved dependency: graphframes#graphframes;0.8.0-spark2.4-s_2.11: not found

mengxr:
See http://apache-spark-user-list.1001560.n3.nabble.com/Bintray-replacement-for-spark-packages-org-td40128.html.

DCameronMauch:
Why was this closed?  Was the issue resolved?

bozhang2820:
As we have notified in the Apache Spark user list (see the link in @mengxr's previous comment), we have spun up a repository service at https://repos.spark-packages.org, and have migrated all existing artifacts (including those for graphframes) to the new repository service. 

Seems that @menglewis is working on fixing this issue in #385.

carwilki:
getting this issue when trying to get get to the repo

```
$ curl -I https://repos.spark-packages.org/graphframes/graphframes/0.8.1-spark2.4-s_2.11/graphframes-0.8.1-spark2.4-s_2.11-sources.jar
HTTP/2 403 
content-type: application/xml
date: Wed, 28 Apr 2021 17:26:26 GMT
[...]

PR #385, opened 2021-04-26
Replace bintray

menglewis:
Bintray is being deprecated which will cause builds to fail because there are dependencies on Bintray: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Also see https://github.com/graphframes/graphframes/issues/384

This PR replaces the usages of Bintray with `repo.scala-sbt.org` which I found from this [issue on SBT](https://github.com/sbt/sbt/issues/6294) about replacing Bintray. I was able to run a build locally with these changes during the April 26 Bintray brownout.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,5 @@
 // You may use this file to add plugin dependencies for sbt.
-resolvers += "Spark Packages repo" at "https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/"
+resolvers += "Spark Packages repo" at "https://repos.spark-packages.org/"
 
 addSbtPlugin("org.spark-packages" %% "sbt-spark-package" % "0.2.6")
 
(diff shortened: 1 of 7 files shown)


project/plugins.sbt
```diff
@@ -1,5 +1,5 @@
 // You may use this file to add plugin dependencies for sbt.
-resolvers += "Spark Packages repo" at "https://dl.bintray.com/spark-packages/maven/"
+resolvers += "Spark Packages repo" at "https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/"
```
bozhang2820, line 2: This should be changed to `https://repos.spark-packages.org/` instead. See: http://apache-spark-user-list.1001560.n3.nabble.com/Bintray-replacement-for-spark-packages-org-td40128.html

menglewis: I'm not sure if I have permissions to re-request a review, but I've made this change and this PR is ready for review again.

Inkvi:
Can this PR be reviewed and merged? With bintray down, it's impossible to include package from the maven repository.

pvillacorta-stratio:
The issue is still there for instance if you need to import it from cloud platforms. E.g. in Google Cloud Dataproc, AFAIK you cannot specify the locaton of the dependencies as it always goes to Maven, therefore it fails to create the cluster with the GraphFrames dependency. 

I really hope the PR can be merged soon :-)

menglewis:
Is there anything else I need to do to get this merged? Happy to help if there's any way I can help.

mengxr:
Merged into master. Thanks!!

PR #443, opened 2023-10-29
make component column string public

lastTwoBrainCells:


rjurney:
@lastTwoBrainCells what is the motivation of this PR?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -163,11 +163,11 @@ object ConnectedComponents extends Logging {
 
   import org.graphframes.GraphFrame._
 
-  private val COMPONENT = "component"
   private val ORIG_ID = "orig_id"
   private val MIN_NBR = "min_nbr"
   private val CNT = "cnt"
   private val CHECKPOINT_NAME_PREFIX = "connected-components"
+  val COMPONENT = "component"
```
SemyonSinchenko, line 170: Why not to add an alternative merhod `run(graph, algorithm, broadcastThreshold, checkpointInterval, intermediateStorageLevel, componentColName)` instead and update the public API? Cause this one is a constant value that works fine for default, but it would be strange to see it as a part of the public API?

SemyonSinchenko:
I'm closing this one due to inactivity.

PR #454, opened 2024-06-03
Fix connected component

WeichenXu123:
Fix connected component algorithm implementation:

The algorithm is described in this paper: https://dl.acm.org/doi/pdf/10.1145/2670979.2670997

Summary of the algorithm: essentially, the algorithm is doing "large-star" and "small-star" transformations over the graph,
until convergence (i.e. the transformation doesn't change graph structure),

```
graph = input_graph  # input graph each node must has a unique and comparable label number
while (! converged) {
   graph = large_star(graph)
   graph = small_star(graph)
}
```

**large star** transformation:
For each node U, finding the neighbor nodes V that has larger label than node U, and finding the node M that has the minimal label in the set of `{*node_U_neighbor_node_set, node_U}`, for every node V, generate an edge from V to M, the generated edges compose of the "large star" transformation output graph.

**small star** transformation:
For each node U, finding the neighbor nodes V that has smaller label than node U, and finding the node M that has the minimal label in the set of `node_U_neighbor_node_set`, for every node V, generate an edge from V to M, and additionally, generate an edge of U to M, the generated edges compose of the "small star" transformation output graph.

image

**Convergence critieria:**

Convergence means large star and small star transformation generates identical output graph, and we can prove that, the output graph has the same component connectivity, and the topology of the graph is the overall graph is a union of disjoint stars, one for each connected component, and each star has a center node with the minimal label in its connected component.

To check if the graph reaches convergence, we can check the sum of all nodes' "minimal label value in set {*neighbor_nodes, self_node}", the sum keeps decreasing for each pass of large-star / small-star transformation, once it stops decreasing, it means the graph reaches convergence, and we can get each node's connectivity component ID by calculating its "minimal label value in set {*neighbor_nodes, self_node}".

So, current graphframe connected component algorithm code has one error in convergence checking, I correct it in this PR.

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -339,18 +339,23 @@ object ConnectedComponents extends Logging {
 
     var prevSum: BigDecimal = _calcMinNbrSum(minNbrs1)
 
+    var lastRoundPersistedDFs = Seq[DataFrame](ee, minNbrs1)
     while (!converged) {
+      var currRoundPersistedDFs = Seq[DataFrame]()
       // large-star step
       // connect all strictly larger neighbors to the min neighbor (including self)
       ee = skewedJoin(ee, minNbrs1, broadcastThreshold, logPrefix)
         .select(col(DST).as(SRC), col(MIN_NBR).as(DST)) // src > dst
         .distinct()
         .persist(intermediateStorageLevel)
+      currRoundPersistedDFs = currRoundPersistedDFs :+ ee
 
       // small-star step
       // compute min neighbors (excluding self-min)
       val minNbrs2 = ee.groupBy(col(SRC)).agg(min(col(DST)).as(MIN_NBR), count("*").as(CNT)) // src > min_nbr
         .persist(intermediateStorageLevel)
+      currRoundPersistedDFs = currRoundPersistedDFs :+ minNbrs2
+
       // connect all smaller neighbors to the min neighbor
       ee = skewedJoin(ee, minNbrs2, broadcastThreshold, logPrefix)
         .select(col(MIN_NBR).as(SRC), col(DST)) // src <= dst
@@ -377,9 +382,11 @@ object ConnectedComponents extends Logging {
       }
 
       ee.persist(intermediateStorageLevel)
+      currRoundPersistedDFs = currRoundPersistedDFs :+ ee
 
       minNbrs1 = minNbrs(ee) // src >= min_nbr
         .persist(intermediateStorageLevel)
+      currRoundPersistedDFs = currRoundPersistedDFs :+ minNbrs1
 
       // test convergence
       val currSum = _calcMinNbrSum(minNbrs1)
@@ -391,6 +398,15 @@ object ConnectedComponents extends Logging {
         prevSum = currSum
       }
 
+      // materialize all persisted DataFrames in current round,
+      // then we can unpersist last round persisted DataFrames.
+      for (persisted_df <- currRoundPersistedDFs) {
+        persisted_df.count()  // materialize it.
+      }
+      for (persisted_df <- lastRoundPersistedDFs) {
+        persisted_df.unpersist()
+      }
+      lastRoundPersistedDFs = currRoundPersistedDFs
       iteration += 1
     }
 
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -314,14 +314,30 @@ object ConnectedComponents extends Logging {
     var converged = false
     var iteration = 1
 
+    def _calcMinNbrSum(minNbrsDF: DataFrame): BigDecimal = {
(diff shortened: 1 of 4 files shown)


src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -356,12 +362,14 @@ object ConnectedComponents extends Logging {
...
-      // test convergence
+      minNbrs1 = minNbrs(ee) // src >= min_nbr
+        .persist(intermediateStorageLevel)
 
+      // test convergence
       // Taking the sum in DecimalType to preserve precision.
       // We use 20 digits for long values and Spark SQL will add 10 digits for the sum.
       // It should be able to handle 200 billion edges without overflow.
-      val (currSum, cnt) = ee.select(sum(col(SRC).cast(DecimalType(20, 0))), count("*")).rdd
+      val (currSum, cnt) = minNbrs1.select(sum(col(MIN_NBR).cast(DecimalType(20, 0))), count("*")).rdd
```
WeichenXu123, line 372: This is the convergence critieria correction.

Ref: Lemma 8 in paper https://dl.acm.org/doi/pdf/10.1145/2670979.2670997

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -313,12 +313,18 @@ object ConnectedComponents extends Logging {
 
     var converged = false
     var iteration = 1
-    var prevSum: BigDecimal = null
+
+    // compute min neighbors (including self-min)
+    var minNbrs1: DataFrame = minNbrs(ee) // src >= min_nbr
+      .persist(intermediateStorageLevel)
+
+    var prevSum: BigDecimal = minNbrs1.select(sum(col(MIN_NBR).cast(DecimalType(20, 0))), count("*")).rdd
```
BenWilson2, line 321: Is there any concern with an overflow by capping this at 20?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -313,12 +313,18 @@ object ConnectedComponents extends Logging {
 
     var converged = false
     var iteration = 1
-    var prevSum: BigDecimal = null
+
+    // compute min neighbors (including self-min)
+    var minNbrs1: DataFrame = minNbrs(ee) // src >= min_nbr
```
BenWilson2, line 318: This makes sense to do the first minimization vector connection reduction outside of the while loop. +1

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -356,12 +362,14 @@ object ConnectedComponents extends Logging {
 
       ee.persist(intermediateStorageLevel)
 
-      // test convergence
+      minNbrs1 = minNbrs(ee) // src >= min_nbr
+        .persist(intermediateStorageLevel)
```
BenWilson2, line 366: Within the termination of the while loop, should we also explicitly call unpersist() in order to remove the cache references to these pseudo-mutable vars that represent an immutable cache state?

BenWilson2:
Let's get the original reporter to confirm with the customer on their data to validate that the convergence correctness is fixed. Let's also be sure to clean up the persistence (as that is something that they complained about) due to the memory cache holding on to those orphan states that were persisted during iterative convergence. 
Once these are done, LGTM!

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -313,12 +313,18 @@ object ConnectedComponents extends Logging {
...
-    var prevSum: BigDecimal = null
+
+    // compute min neighbors (including self-min)
+    var minNbrs1: DataFrame = minNbrs(ee) // src >= min_nbr
+      .persist(intermediateStorageLevel)
+
+    var prevSum: BigDecimal = minNbrs1.select(sum(col(MIN_NBR).cast(DecimalType(20, 0))), count("*")).rdd
+      .map { r =>
+        (r.getAs[BigDecimal](0), r.getLong(1))
+      }.first()._1
```
serena-ruan, line 324: Can we make this a function and reuse it here and below?

serena-ruan (commented):
Do we have test to validate the algorithm is correct?

WeichenXu123:
> Do we have test to validate the algorithm is correct?

I will generate some random graph to test.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
WeichenXu123: sure. good idea

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
WeichenXu123: I will revisit and fix the persist/unpersist operations

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
WeichenXu123: added verification.

minerjaime:
This is great!

Issue #458, opened 2024-09-20
ConnectedComponents Orphaning Cached Dataframes


james-willis:
Running connected components leaves dataframes cached in the spark runtime which can take up a substantial amount of memory when it is run many times. Specifically, the dfs in `lastRoundPersistedDFs` do not get cleaned up after `convergence == true`

Since the cached dataframe is not the one returned by the run method, the caller is unable to unpersist them without clearing all persisted dataframes.

SemyonSinchenko:
I think it was done in #552

PR #459, opened 2024-09-20
Do not orphan out of scope persisted dataframes in ConnectedComponent…

james-willis:
…s.run function

Fix for https://github.com/graphframes/graphframes/issues/458

### DIFF
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -440,10 +440,20 @@ object ConnectedComponents extends Logging {
 
       logInfo(s"$logPrefix Connected components converged in ${iteration - 1} iterations.")
 
-      logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
-      vv.join(ee, vv(ID) === ee(DST), "left_outer")
-        .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
-        .select(col(s"$ATTR.*"), col(COMPONENT))
+    logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
+    val output = vv.join(ee, vv(ID) === ee(DST), "left_outer")
+      .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
+      .select(col(s"$ATTR.*"), col(COMPONENT)).persist(intermediateStorageLevel)
+
+    // materialize the output DataFrame
+    output.count()
+
+    // clean up persisted DFs
+    for (persisted_df <- lastRoundPersistedDFs) {
+      persisted_df.unpersist()
+    }
+
+    output
     } finally {
       // Restore original AQE setting
       spark.conf.set("spark.sql.adaptive.enabled", originalAQE)
--- a/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -17,17 +17,16 @@
 
 package org.graphframes.lib
 
+import org.apache.hadoop.fs.Path
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.DecimalType
+import org.apache.spark.sql.{Column, DataFrame}
+import org.apache.spark.storage.StorageLevel
+import org.graphframes.{GraphFrame, Logging}
+
 import java.io.IOException
 import java.math.BigDecimal
 import java.util.UUID
-
-import org.graphframes.{GraphFrame, Logging}
-
-import org.apache.hadoop.fs.{FileSystem, Path}
-import org.apache.spark.sql.{Column, DataFrame}
-import org.apache.spark.sql.functions._
-import org.apache.spark.sql.types.DecimalType
-import org.apache.spark.storage.StorageLevel
 
 /**
  * Connected components algorithm.
@@ -440,20 +439,22 @@ object ConnectedComponents extends Logging {
 
       logInfo(s"$logPrefix Connected components converged in ${iteration - 1} iterations.")
 
-    logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
-    val output = vv.join(ee, vv(ID) === ee(DST), "left_outer")
-      .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
-      .select(col(s"$ATTR.*"), col(COMPONENT)).persist(intermediateStorageLevel)
+      logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
+      val output = vv
+        .join(ee, vv(ID) === ee(DST), "left_outer")
+        .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
+        .select(col(s"$ATTR.*"), col(COMPONENT))
+        .persist(intermediateStorageLevel)
 
-    // materialize the output DataFrame
-    output.count()
+      // materialize the output DataFrame
+      output.count()
 
-    // clean up persisted DFs
-    for (persisted_df <- lastRoundPersistedDFs) {
-      persisted_df.unpersist()
-    }
+      // clean up persisted DFs
+      for (persisted_df <- lastRoundPersistedDFs) {
+        persisted_df.unpersist()
+      }
 
-    output
+      output
     } finally {
       // Restore original AQE setting
       spark.conf.set("spark.sql.adaptive.enabled", originalAQE)
--- a/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
+++ b/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
@@ -17,19 +17,19 @@
 
 package org.graphframes.lib
 
-import java.io.IOException
-
-import scala.reflect.ClassTag
-import scala.reflect.runtime.universe.TypeTag
-
-import org.apache.spark.sql.{DataFrame, Row}
+import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
+import org.apache.spark.sql.execution.columnar.InMemoryTableScanExec
 import org.apache.spark.sql.functions.{col, lit}
 import org.apache.spark.sql.types.DataTypes
+import org.apache.spark.sql.{DataFrame, Row}
 import org.apache.spark.storage.StorageLevel
-
-import org.graphframes._
 import org.graphframes.GraphFrame._
+import org.graphframes._
 import org.graphframes.examples.Graphs
+
+import java.io.IOException
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe.TypeTag
 
 class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkContext {
 
@@ -253,6 +253,20 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
     }
   }
 
+  test("uses intermediate caches") {
+    val cc = Graphs.friends.connectedComponents
+    val components = cc.run()
+
+    val count = components.queryExecution.executedPlan
+      .toString()
+      .sliding("InMemoryRelation".length)
+      .count(window => window == "InMemoryRelation")
+
+    // 17 number derived from when output.count() call is present in the run method
+    assert(count == 17)
+    components.unpersist(blocking = true)
+  }
+
   test("not leaking cached data") {
     val priorCachedDFsSize = spark.sparkContext.getPersistentRDDs.size
 
@@ -260,7 +274,6 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
     val components = cc.run()
 
     components.unpersist(blocking = true)
-
     assert(spark.sparkContext.getPersistentRDDs.size === priorCachedDFsSize)
   }
 
--- a/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
+++ b/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
@@ -253,6 +253,17 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
     }
   }
 
+  test("not leaking cached data") {
+    val priorCachedDFsSize = spark.sparkContext.getPersistentRDDs.size
+
+    val cc = Graphs.friends.connectedComponents
+    val components = cc.run()
+
+    components.unpersist(blocking = true)
+
+    assert(spark.sparkContext.getPersistentRDDs.size === priorCachedDFsSize)
+  }
+
   private def assertComponents[T: ClassTag: TypeTag](
       actual: DataFrame,
       expected: Set[Set[T]]): Unit = {


james-willis:
@rjurney can you take a look?

rjurney:
@james-willis thanks for this, will look it over and test it

rjurney:
@SemyonSinchenko @bjornjorgensen could one of you plz take a look at this? Is this involved in the recent issue @SauronShepherd investigated in Spark?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -416,11 +416,21 @@ object ConnectedComponents extends Logging {
       }
 
       logInfo(s"$logPrefix Connected components converged in ${iteration - 1} iterations.")
-
+      
       logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
       vv.join(ee, vv(ID) === ee(DST), "left_outer")
         .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
-        .select(col(s"$ATTR.*"), col(COMPONENT))
+        .select(col(s"$ATTR.*"), col(COMPONENT)).persist(intermediateStorageLevel)
```
SemyonSinchenko, line 423: I think it is a user-facing change because from now an output of `ConnectedComponents.run()` is a persisted `DataFrame`. It may create problems in some cases, for example, if users are calling it multiple time or for multiple small subgraphs. In that case persisted `DataFrame`'s will be in memory forever. I think we should at least change the documentation of the function to explicitly warning an end user that an output is persisted from now. As a better solution I would like to move it to one of the further releases with breaking changes (like 1.0) and disable persisting of the output for now.

SemyonSinchenko: Can we, maybe, only unpersist the temporary cache to fix the bug and make persisting of the output optional or disabled until the new big release?

james-willis: > It may create problems in some cases, for example, if users are calling it multiple time or for multiple small subgraphs.

This is already the kind of issue users are encountering today because of the orphaned intermediate dataframes. At least with this change the dataframe that is persisted is one that is available in the user's scope. This way the user can unpersist what is persisted by graphframes without nuking their entire cache.

I view this as only an improvement in regards to this kind of memory leak issue caused by this caching bug.

> Can we, maybe, only unpersist the temporary cache to fix the bug and make persisting of the output optional or disabled until the new big release?

If you unpersist the temporary cache before you cache the output, you will have a cache misses when you materialize the output dataframe. Because of the [checkpointing code](https://github.com/graphframes/graphframes/blob/e4a2024f2d1271e6af973f4b25b7fef06f66ae00/src/main/scala/org/graphframes/lib/ConnectedComponents.scala#L373) this isn't as bad as having no persisting at all, but can be quite expensive on larger graphs. TBH I don't have a good sense of how expensive your proposed solution would be in terms of execution time.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -416,11 +416,21 @@ object ConnectedComponents extends Logging {
...
+
+      // materialize the output DataFrame
+      output.count()
```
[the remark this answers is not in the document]
james-willis: looks like i screwed this up when I merged main back into my fork. Let me fix

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
james-willis: We actually discovered this bug in sedona when our unit tests were going OOM in a feature that uses graphframes' connected components 

See here: https://github.com/apache/sedona/pull/1589

SauronShepherd: I wrote an [article](https://medium.com/towards-data-engineering/apache-spark-wtf-i-like-it-when-a-plan-comes-together-part-ii-dc59def302b3) discussing this issue. The method tracks every persisted DataFrame, so I don’t see how a memory leak could occur because of that. The OOM issue I’ve been analyzing is not directly related to persisting DataFrames but rather to how Spark constantly generates string representations of execution plans. This behavior is independent of unpersisting DataFrames and is not resolved by doing so. If the bug in Sedona gets fixed by disabling AQE, most probably you're dealing with the same global issue, and it's not a bug of Graphframes, but a bug of Spark [SPARK-50992](https://issues.apache.org/jira/browse/SPARK-50992). I've even create a [PR](https://github.com/apache/spark/pull/49724)

james-willis: @SauronShepherd Is there something wrong with the test in this PR?

If you run that test against HEAD of master it'll fail, showing that the query is orphaning persisted Dataframes.

I don't discount that AQE's string serialization issue is real and exerts memory pressure on the driver, but it doesn't explain why the nth execution of connected components goes OOM but not all those m executions where m < n. 

To be fully transparent I only skimmed your article and the emails from the spark dev mailing list.

james-willis: > The method tracks every persisted DataFrame, so I don’t see how a memory leak could occur because of that. 

This is the crux of this whole PR. In master, the function doesn't unpersist the last round of persisted dataframes. 

Thats why I wrote a test and raised a PR to fix it.

SauronShepherd: Ah, I see. You're right—the last round of persisted DataFrames won't get unpersisted. However, that still doesn't explain an OOM. Spark uses an LRU algorithm for caching/persisting, so persisted DataFrames should be evicted as soon as new ones are persisted. The only issue with not unpersisting DataFrames is that it could introduce a very small memory leak on the driver side. This leak only becomes problematic in the very long run. For example, two years ago, I encountered an issue in a streaming application where I forgot to unpersist a DataFrame. Over time, the memory usage on the driver slowly increased, and after more than eight hours of continuously executing microbatches, an OOM was thrown. That’s the full extent of the memory leak.

SauronShepherd: I guess the last round of DataFrames is not being unpersisted because they rely on previously persisted DataFrames that have already been unpersisted. The only way to prevent Spark from recomputing all the operations when an action is performed (outside the method) is to keep those DataFrames persisted. Does that make sense?

SauronShepherd: I'll try to take a look at the Sedona issue today because it has sparked my curiosity. 😉

james-willis: > That’s the full extent of the memory leak.

Yes my coworker said theres some element here where the fact there's broadcast joins being cached exacerbates the memory used. I didnt totally follow but disabling broadcast joins for the tests in sedona allowed the tests to pass in the low memory test environment in the CI/CI. I bet the AQE issue also helps realize the memory pressure issue in the tests.

> The only way to prevent Spark from recomputing all the operations when an action is performed (outside the method) is to keep those DataFrames persisted. Does that make sense?

Agreed. thats why I proposed the fix in this PR; the only way I can think to leverage persist while making it so the user can unpersist is to have that output DF be the only one that is persisted past the end of the function call.

> I'll try to take a look at the Sedona issue today because it has sparked my curiosity. 😉

Since it arose only within the context of that PR it wont be the most well documented thing. you can see the different in the CI actions that pass for 5f176ea and 8aeaee1 (ignore the R test failures here!). Disabling broadcast joins save enough ram to get the tests to pass

SemyonSinchenko: > This is already the kind of issue users are encountering today because of the orphaned intermediate dataframes. At least with this change the dataframe that is persisted is one that is available in the user's scope. This way the user can unpersist what is persisted by graphframes without nuking their entire cache.

@james-willis Can we update the documentation in that case? Just add a line to the docstring of the public `run` that the output is persisted DataFrame and user should care of unpersisting it? Thanks!

SauronShepherd: > > my coworker said theres some element here where the fact there's broadcast joins being cached exacerbates the memory used. I didnt totally follow but disabling broadcast joins for the tests in sedona allowed the tests to pass in the low memory test environment in the CI/CI. 

OMG, you're spicing things up! Now looking into the Sedona issue seems even more interesting!

>> Agreed. thats why I proposed the fix in this PR; the only way I can think to leverage persist while making it so the user can unpersist is to have that output DF be the only one that is persisted past the end of the function call.

But then ... we'd still persisting a DF inside the method. Is that a fix or just another workaround?
I'm thinking ... maybe disabling the AQE temporarily doesn't really fix the problem, because it may only apply to the actions taken place inside the method. Once the AQE is enabled again and another action is performed on the DF, the OOM raises again. Haven't test it yet, but it has some sense in my head.

>> Since it arose only within the context of that PR it wont be the most well documented thing. 

Don't worry about that. Do you mind if I send you an email if I have some doubt's/problems reproducing the issue?

SauronShepherd: > Can we update the documentation in that case? Just add a line to the docstring of the public `run` that the output is persisted DataFrame and user should care of unpersisting it? Thanks!

The output DF had already persisted parts inside its plan, but I think it's a good warning.

james-willis: > Can we update the documentation in that case?

I think this is a good idea regardless of what we decide. 

> Now looking into the Sedona issue seems even more interesting!

here's some messages from the aforementioned coworker:
> The reason why we are running into OOM is that: the broadcasted relations won't be automatically unpersisted, since the broadcast joins are part of the query plans of cached dataframes.
When GC happens on the driver, unreferenced broadcast variables will be unpersisted. However, cached dataframes containing broadcast joins hold strong references to broadcast variables, which prevent them from being deallocated.
Caching prevents broadcasted variables from being reclaimed, I'm not sure if it is a bug of spark or not. To workaround this, we can try disabling auto broadcast join or calling spark.catalog.clearCache() after each dbscan run.


> we'd still persisting a DF inside the method. Is that a fix or just another workaround?

With this fix, the only thing that is left persisted from inside the method is the dataframe that is returned, so it is possible for the user to unpersist it without `spark.catalog.clearCache()`. I would say this is definitely more fixed than persisting dfs that the caller cannot unpersist individually.
Is it optimal? Well- it would be optimal that spark/gf is smarter than this and I never have to think about calling persist. Until we reach that point, using persist in general is a trade off of usability and performance.

SemyonSinchenko: > With this fix, the only thing that is left persisted from inside the method is the dataframe that is returned, so it is possible for the user to unpersist it without spark.catalog.clearCache().

Sounds good for me!

james-willis: > Do you mind if I send you an email if I have some doubt's/problems reproducing the issue?

I do not mind.

SauronShepherd: I haven't been able to reproduce the issue in Sedona. The GeoStatsSuite tests (the ones which use ConnectedComponents) worked fine, even after enabling the "sedona.join.autoBroadcastJoinThreshold" property.  The heap went up to around 4GB, but all the 15 tests passed.

james-willis: The gh hosted runner should have 16GB of ram, [That probably hasn't changed since September](https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/)

You should enable both the sedona and spark autobroadcastjointhresholds.

I never was able to get this behavior on my M3 Mac (18GB ram). Only in the CI did it occur. 

Since the logs from that run have expired its a bit hard to remember the details of the failure :( 

@Kontinuation do you remember anything more specific about those test failures?

Kontinuation: I remembered that the failed tests are python tests in [python/tests/stats/test_dbscan.py](https://github.com/apache/sedona/blob/master/python/tests/stats/test_dbscan.py). The spark driver process launched by pyspark will have only 1GB maximum heap space by default, regardless of the amount of memory we have on the runner.

The version of Spark we were using was Spark 3.5.0. I've just reran the test using 3.5.0 and haven't seen the AQE problem, but later versions such as Spark 3.5.3 has the AQE problem, which is quite interesting.

SauronShepherd: Setting the Spark master to local[4] (instead of utilizing all 16 cores on my laptop) increases the maximum heap size to 6GB, but the test still passes. However, I noticed that the number of SparkSession instances reached 117. Each time a DF is persisted, a new SparkSession is created, which is then destroyed when the DF is unpersisted.

The memory leak in GraphFrames, caused by not unpersisting the last round of DFs, isn't that big. However, I realised that Sedona is using graphframes-0.8.3-spark3.4-s_2.12, which lacks the unpersist loop. This issue was addressed in a fix introduced in 2024, while the ConnectedComponents class dates back to 2023.

SauronShepherd: > Heap dump also showed that the memory usage was dominated by `UnsafeHashRelation`s originated from BroadcastExchange. We worked around this problem by disabling broadcast but we'd still like to know the root cause of this problem.

Thanks to your example, I've been able to reproduce the issue, but only for Spark versions 3.5.x. In Spark 3.4.4, the number of `UnsafeHashedRelation` instances seems to increase only by 1 per iteration. 

Also, in Spark 3.5.x, disabling the AQE also fixes the problem. Looks like something changed from 3.4 to 3.5 and now, when a persisted DF with broadcast is not unpersisted, things get somehow out of control ...

rjurney: @SauronShepherd so what do we do about this issue?

SauronShepherd: It should be fixed with the latest GraphFrames changes (temporarily disabling AQE). Also, I'd like to submit my ConnectedComponents changes (a new experimental implementation) soon.

rjurney: Okay, sounds good.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -435,10 +435,20 @@ object ConnectedComponents extends Logging {
...
-      vv.join(ee, vv(ID) === ee(DST), "left_outer")
-        .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
-        .select(col(s"$ATTR.*"), col(COMPONENT))
+    logInfo(s"$logPrefix Join and return component assignments with original vertex IDs.")
+    val output = vv.join(ee, vv(ID) === ee(DST), "left_outer")
+      .select(vv(ATTR), when(ee(SRC).isNull, vv(ID)).otherwise(ee(SRC)).as(COMPONENT))
+      .select(col(s"$ATTR.*"), col(COMPONENT)).persist(intermediateStorageLevel)
+
+    // materialize the output DataFrame
+    output.count()
```
ericsun95, line 444: Nit, can we just persist with eager=true instead of calling an action? I think a lot code were written with old spark and we can later simplify a lot.

james-willis: Hi Peiyuan! I think this PR isnt getting merged so I'm going to abandon it.

rjurney: @SauronShepherd verify this doesn't need merging?

SauronShepherd:
I've just opened a related PR (#552) with the counts removed. Could we close this PR and, if nobody has any objections, approve the other one (#552)?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
james-willis: Since we intend to merge one of these PRs, I've copied my comment over from the other PR:

> I believe persist is lazy and does not offer an eager flag. Will this code actually wind up using the cached dataframes if we dont cache the output df before we unpersist the child dataframes?

james-willis: Tried to add a test to show this.

SemyonSinchenko:
Closed because #552 was merged.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
ericsun95: > Since we intend to merge one of these PRs, I've copied my comment over from the other PR:
> 
> 
> 
> > I believe persist is lazy and does not offer an eager flag. Will this code actually wind up using the cached dataframes if we dont cache the output df before we unpersist the child dataframes?

yeah,the checkpoint function has eager option but not the persist.

PR #462, opened 2024-11-11
Moved from deprecated MaxPermSize to MaxMetaspaceSize to make build work

rjurney:
The `build/sbt assembly` build crashes on OpenJDK 21 due to a `MaxPermSize` option in sbt build configuration that was removed in Java 9. Updated it to use `MaxMetaspaceSize` which isn't the same but is the recommended replacement. Build gets through this step once I do so.


rjurney:
Used another branch...

PR #471, opened 2024-11-20
Upgrade `github actions`

bjornjorgensen:
This will upgrade actions/checkout and actions/cache from version 3 to version 4

Release info https://github.com/actions/cache  and https://github.com/actions/checkout


rjurney:
> @rjurney start with updating the files for github workflows, before somethings like #470

I appreciate the PR. I am testing it now.

PR #474, opened 2024-12-26
Use vertices SparkSession

Nassizouz:
The previous version of the `GraphFrame` class used the vertices SQLContext, when the code was changed to use SparkSession the thread's session was taken, this change can break old working code if the GraphFrame is initialized in a thread that didn't initialized a spark session.
This PR uses the vertices SparkSession so it will work regardless to whether a spark session was initialized on the current thread

### DIFF
--- a/python/graphframes/graphframe.py
+++ b/python/graphframes/graphframe.py
@@ -63,7 +63,7 @@ class GraphFrame(object):
     def __init__(self, v, e):
         self._vertices = v
         self._edges = e
-        self._spark = SparkSession.getActiveSession()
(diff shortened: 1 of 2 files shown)


rjurney:
@Nassizouz is there any way you can create a simple unit test for this?

Nassizouz:
@rjurney Added a simple unittest

rjurney:
@Nassizouz Testing this now... thanks for the bugfix :) Keep it coming!

I need to merge my fix for Java 8 before I can test this in Java 8, one minute...

PR #476, opened 2025-01-11
Fix the bug I introduced previously when adapting `build.sbt` for JDK 11

rjurney:
JDK 8 support is adequate for now, I believe it is still the recommended version for Spark.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,8 @@ project/plugins/project/
 # intellij
 .idea/
 
+# VSCode
+.vscode
+
 # Mac
 *.DS_Store
--- a/build.sbt
+++ b/build.sbt
@@ -66,9 +66,7 @@ fork in Test := true
 javaOptions in Test ++= Seq(
   "-Xmx2048m",
   "-XX:ReservedCodeCacheSize=384m",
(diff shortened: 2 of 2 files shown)


bjornjorgensen:
for spark 4.0 we need java 17 and 21

rjurney:
> for spark 4.0 we need java 17 and 21

Okay, I will test again but this seemed to break Java 8, so I thought of pulling it. I will check again.

rjurney:
@bjornjorgensen yeah this breaks Java 0.8, which I think most Spark users run. I guess we should conditionally add it based on the JVM version?

```python
Unrecognized option: --add-opens=java.base/sun.nio.ch=ALL-UNNAMED
```

Kimahriman:
Spark also includes `-XX:+IgnoreUnrecognizedVMOptions` so it works for both Java versions

bjornjorgensen:
spark branch 3.5 only support and build on java version 11 and 17 
https://github.com/apache/spark/actions/runs/12743044791/job/35512274672 
image


I have with a lot of help from AI tested it to run on java 8, 11 and 17 https://github.com/bjornjorgensen/graphframes/commits/java11/ you can copy what you like but.. its made just to test and it is written by AI

PR #480, opened 2025-01-16
Add Type hints for Python code 

bjornjorgensen:
Add type hints for python code that works for python 3.9

### DIFF
--- a/python/graphframes/examples/belief_propagation.py
+++ b/python/graphframes/examples/belief_propagation.py
@@ -16,6 +16,7 @@
 #
 
 import math
+from typing import Union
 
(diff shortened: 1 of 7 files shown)


python/graphframes/examples/belief_propagation.py
```diff
@@ -27,7 +28,7 @@
 __all__ = ['BeliefPropagation']
 
 
-class BeliefPropagation(object):
+class BeliefPropagation:
```
bjornjorgensen, line 31: https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object

rjurney (approved):

PR #507, opened 2025-02-10
chore: add scalafmt and re-format all the code

### LINKED ISSUE #500: chore: scalafmt file and linting rules
At the moment there is no `.scalafmt` file and no sbt plugin to check the scala code. I do not see any alternatives (like scalastyle or scalafix) too.

An example of `.scalafmt` file: https://github.com/apache/spark/blob/master/dev/.scalafmt.conf
SBT plugin: https://scalameta.org/scalafmt/docs/installation.html#sbt

Checking the code style on each PR is a good practice that allows to simplify the maintenance.

SemyonSinchenko:
This PR closes #500 

Changes:
- `.scalafmt.conf` is added with content exactly like in the Apache Spark Project
- `scalafmt` plugin is added to the plugins
- all the code is reformatted according to the rules
- checking the formatting is added to the compile stage
- `.metals` is added to the `.gitignore`

P.S. Metals is the most popular LSP for Scala that is used in VSCode, Vim / NeoVim, Emacs, Sublime, etc. I think it is nice to have it in `.gitignore` because it simplify the working with git and reduce a chance of accidentally pushing something not related to the project itself.
P.P.S. All the changes in the scala code is related to formatting only! There are no actual changes of the logic.


SemyonSinchenko:
@rjurney @bjornjorgensen Hello! Can you review it, please? It may look huge but there are no changes in the code, only formatting of the code. Thanks!

.scalafmt.conf
```diff
@@ -0,0 +1,32 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
bjornjorgensen, line 2: Is it a ASF project?

SemyonSinchenko: Exactly this file is from the Apache Spark repo, so I think we should preserve the header. Overall, this header is very nice to have because it will also simplify the donation of the project to ASF or making it a part of the Apache Spark.

bjornjorgensen: yes, but as long as this is not an ASF project I don't think we can say that it is.

SemyonSinchenko: Ok. I removed the header.

build.sbt
```diff
@@ -39,6 +39,9 @@ lazy val root = (project in file("."))
     // Modern way to set Scala options
     Compile / scalacOptions ++= Seq("-deprecation", "-feature"),
 
+    // Scalafmt
+    scalafmtOnCompile := true,
```
bjornjorgensen, line 43: I think we should remove this. 
It will format every PR and contributes have to sync the PR to local every time.

bjornjorgensen: In spark it only check but write a error msg https://github.com/apache/spark/blob/dd153307cb9735fd05a41124eca2a136f40f3b3f/dev/lint-scala#L41

bjornjorgensen: What do you think @SemyonSinchenko ?

SemyonSinchenko: I removed it.

bjornjorgensen: but how are we going to know if the code is in style then?

SemyonSinchenko: > but how are we going to know if the code is in style then?

We should run `root/scalafmtCheckAll` as part of the CI. I can do it as a part of the current PR or I can open a new one. The idea of calling `scalafmtOnCompile := true` was exactly to avoid changes in CI and forcing `scalafmt` on running test or assembly.

bjornjorgensen: ok.

bjornjorgensen (approved):
LGTM

rjurney:
@SemyonSinchenko oh, one thing - are these the same settings as Spark? The plan is to eventually merge GraphFrames into Spark main, so consistency with their settings would be nice.

SemyonSinchenko:
> @SemyonSinchenko oh, one thing - are these the same settings as Spark? The plan is to eventually merge GraphFrames into Spark main, so consistency with their settings would be nice.

I copied the whole `.scalafmt` from the Apache Spark repository. So yes, these settings are the same.

SemyonSinchenko:
@rjurney Can we merge this one? It allows to apply `scalafmt` to all other pending PRs.

bjornjorgensen:
@rjurney are there anything wrong with this PR?

rjurney:
Merging... @bjornjorgensen nothing wrong :)

PR #512, opened 2025-02-16
Various build / test / package imrovements: Use PyTest, build Python package, etc.

rjurney:
I did some improvements in the project's build and testing in the course of #473 that I have split out into this PR.

PR #511, opened 2025-02-16
3 of 3: Documentation cleanup and update. Added a motif finding tutorial.

rjurney:
This is a documentation only change, the docs part of #473. The docs build and look okay when I serve them.

### What changes were proposed in this pull request?

I added a network motif tutorial and updated / cleaned the docs generally. I added the new google group and the new logo to the images folder.

### Why are the changes needed?

The docs were wrong in a few ways and there weren't even instructions in the README, which stopped new users from onboarding.

PR #473, opened 2024-12-25
Adding motif finding tutorial using the stats.meta.stackexchange.com data dump

rjurney:
This PR makes the following additions to create a tutorial on motif finding using [`stats.meta.stackexchange.com` data dump at the internet archive](https://archive.org/details/stackexchange). Teaching the concepts behind this powerful tool will drive increased adoption of GraphFrames.

* A new tutorial on motif finding in `docs/motif-tutorial.md`
* Code for the demo in `python/graphframes/examples` - `download.py`, `xml_to_parquet.py`, `graph.py` and `motif.py`
* A new data folder `python/graphframes/examples/data`

This code was originally written by myself under the [MIT License](https://github.com/Graphlet-AI/graphml-class/blob/main/LICENSE) for a class at [Connected Data London 2024](https://2024.connected-data.london) called [Full Stack Graph Machine Learning](https://2024.connected-data.london/talks/full-stack-graph-machine-learning/). It can be found at https://github.com/Graphlet-AI/graphml-class/tree/main/graphml_class/stats

PR #518, opened 2025-02-17
2 of 3: New minimized PR for a Python tutorial module graphframes.tutorial

rjurney:
This is a sub-PR of the monster PR #473. It is the code that corresponds to #511. It needs to get merged after #512 and before the docs PR #511.

These changes do the following:

1. Allows users to download the Stack Exchange data dump via a CLI at `graphframes.tutorials.download`. [Thought: this `Click` usage can be the basis for future CLI commands for a `graphframes` command? Just an idea.]
2. Convert the XML to a Parquet file `graphframes.tutorials.stackexchange`
3. Build a test knowledge graph out of the data dump `graphframes.tutorials.stackexchange`. No longer requires case sensitivity for `id` / `Id` fields.
4. Run some motifs on that knowledge graph `graphframes.tutorials.motif` - this is here to provide for future unit testability of tutorials and as a Github browsable reference that matches the Motif Finding tutorial in #511.

In addition:

1. The Stack Exchange knowledge graph dataset `Nodes.parquet` and `Edges.parquet` this PR creates can be wired into the unit tests for a more realistic setting in a near future PR by me. We could put the `python/graphframes/tutorials/data/` folder under `python/data` or `python/graphframes/data` to accommodate this. We need a real dataset for our unit tests, I don't have confidence in changes to algorithms like connected components or PageRank without real data and known outcomes.

These changes are needed to make the docs in #511 work. Otherwise that PR's new Motif Finding Tutorial won't work. Merge me first :)


rjurney:
@bjornjorgensen @SemyonSinchenko @SauronShepherd @WeichenXu123 okay guys, this is the second PR after #512 - it is based on that branch as it uses poetry to define a `tutorials` group you install with `poetry install --with tutorials]` or once it is up on PyPi via `pip install graphframes[tutorials]`.

rjurney:
@SemyonSinchenko can you take a look at this one?

python/pyproject.toml
SemyonSinchenko, line 1: Why not to add CLI here (download)?
FYI: https://python-poetry.org/docs/pyproject/#scripts

python/graphframes/tutorials/download.py
```diff
@@ -0,0 +1,71 @@
...
+        # Extract if requested
+        if extract:
+            click.echo("Extracting archive...")
+            output_dir = f"{subdomain}.stackexchange.com"
+            with py7zr.SevenZipFile(archive_path, mode="r") as z:
+                z.extractall(path=os.path.join(data_dir, output_dir))
+            click.echo(f"Extraction complete: {output_dir}")
+
+    except requests.exceptions.RequestException as e:
+        click.echo(f"Error downloading archive: {e}", err=True)
```
SemyonSinchenko, line 63: Let's maybe try for a couple of times in case of network errors? 2-3 should be enough

python/graphframes/tutorials/motif.py
```diff
@@ -0,0 +1,204 @@
...
+#
+
+import pyspark.sql.functions as F
+from pyspark import SparkContext
+from pyspark.sql import DataFrame, SparkSession
+
+from graphframes import GraphFrame
+
+# Initialize a SparkSession
+spark: SparkSession = SparkSession.builder.appName("Stack Overflow Motif Analysis").getOrCreate()
```
SemyonSinchenko, line 16: What do you think about passing checkpoint dir during the session init and avoid at all any usage of SparkContext (in PySpark docs it is recommended to use SparkSession instead of SparkContext)?

python/graphframes/tutorials/stackexchange.py
```diff
@@ -0,0 +1,595 @@
...
+    # Remove < and > and split into array
+    return re.findall(r"<([^>]+)>", tags)
+
+
+#
+# Initialize a SparkSession with case sensitivity
+#
+
+spark: SparkSession = SparkSession.builder.appName("Stack Exchange Graph Builder").getOrCreate()
+sc = spark.sparkContext
```
SemyonSinchenko, line 51: See my comment in motif about SparkContext vs SparkSession

python/graphframes/tutorials/stackexchange.py
```diff
@@ -0,0 +1,595 @@
...
+
+# Remove the _ prefix from field names
+badges_df = remove_prefix(badges_df).withColumn("Type", F.lit("Badge"))
+
+
+#
+# Form the nodes from the UNION of posts, users, votes and their combined schemas
+#
+
+all_cols: List[Tuple[str, T.StructField]] = list(
```
SemyonSinchenko, line 242: Let's add `from __future__ import annotations` and use `list[tuple[str, T.StructField]]` instead?

SemyonSinchenko (commented):
@rjurney LGTM overall, I left a few minor comments.

rjurney:
All comments addressed... when it passes build checks, I'm gonna merge :)

PR #524, opened 2025-02-25
Convert nose / unittest tests to pytest

rjurney:

I converted the `unittest`/`nose` tests to `pytest`.

This is important because `nose` does not support Python 3.10, which means we can't either and still run unit tests.


python/graphframes/tests.py
SemyonSinchenko, line 1: Let's move it to python/tests/test.py?

python/graphframes/tests.py
```diff
@@ -15,63 +15,70 @@
...
 
     @classmethod
     def createSparkContext(cls):
-        cls.sc = sc = SparkContext('local[4]', "GraphFramesTests")
+        cls.conf = SparkConf().setAppName("GraphFramesTests")
+        cls.conf.set(
+            "spark.submit.pyFiles",
+            os.path.abspath("python/dist/graphframes-0.8.4-py3-none-any.whl"),
+        )
+        cls.sc = SparkContext(master="local[4]", appName="GraphFramesTests", conf=cls.conf)
```
SemyonSinchenko, line 78: Can we avoid any usages of the SparkContext at all and rely only on a SparkSession? It allows smoothly test both spark classic and spark connect. You can see how it is done in #506

python/graphframes/tests.py
```diff
@@ -81,39 +88,42 @@ def stopSparkContext(cls):
...
-
-    @classmethod
-    def setUpClass(cls):
-        # Small tests run much faster with spark.sql.shuffle.partitions = 4
-        cls.spark = SparkSession(GraphFrameTestUtils.sc).builder.config('spark.sql.shuffle.partitions', 4).getOrCreate()
+@pytest.fixture(scope="class")
+def spark_session():
+    # Create a SparkSession with a smaller number of shuffle partitions.
+    spark = (
+        SparkSession(GraphFrameTestUtils.sc)
```
SemyonSinchenko, line 114: We can just explicitly pass all the same confs (app name, etc.) to the `SparkSession.builder`

python/graphframes/tests.py
```diff
@@ -15,63 +15,70 @@
...
             pass
         return version_info
 
     @classmethod
     def createSparkContext(cls):
-        cls.sc = sc = SparkContext('local[4]', "GraphFramesTests")
+        cls.conf = SparkConf().setAppName("GraphFramesTests")
+        cls.conf.set(
+            "spark.submit.pyFiles",
+            os.path.abspath("python/dist/graphframes-0.8.4-py3-none-any.whl"),
```
SemyonSinchenko, line 76: Why do we need it? If the project is installed via `poetry install` to the same venv, like pyspark, it will be here anyway.

python/graphframes/tests.py
```diff
@@ -204,74 +224,93 @@ def test_dropIsolatedVertices(self):
...
+            .withVertexColumn(
+                "rank",
+                F.lit(1.0 / numVertices),
+                F.coalesce(Pregel.msg(), F.lit(0.0)) * F.lit(1.0 - alpha)
+                + F.lit(alpha / numVertices),
+            )
+            .sendMsgToDst(Pregel.src("rank") / Pregel.src("outDegree"))
+            .aggMsgs(F.sum(Pregel.msg()))
             .run()
+        )
```
SemyonSinchenko, line 279:
```suggestion
        pregel = graph.pregel
        ranks = (
            pregel.setMaxIter(5)
            .withVertexColumn(
                "rank",
                F.lit(1.0 / numVertices),
                F.coalesce(pregel.msg(), F.lit(0.0)) * F.lit(1.0 - alpha)
                + F.lit(alpha / numVertices),
            )
            .sendMsgToDst(pregel.src("rank") / pregel.src("outDegree"))
            .aggMsgs(F.sum(pregel.msg()))
            .run()
        )
```

Because with support of SparkConnect there will be two possible types (Pregel classic and Pregel connect)

rjurney:
@SemyonSinchenko I'm starting my review of your PR now, it will take a few days... but do you mind if we get this one in first?

rjurney:
@SemyonSinchenko trying to finish this as quickly as possible, almost got it done... waiting to see if tests pass :)

python/graphframes/tests.py
rjurney: @SemyonSinchenko do you mean this? Can I do the equivalent in Python?

```scala
      case MethodCase.CONNECTED_COMPONENTS => {
        val cc = apiMessage.getConnectedComponents
        graphFrame.connectedComponents
          .setAlgorithm(cc.getAlgorithm)
          .setCheckpointInterval(cc.getCheckpointInterval)
          .setBroadcastThreshold(cc.getBroadcastThreshold)
          .run()
      }
```

rjurney: @SemyonSinchenko it looks like this code was removed, so no longer an issue?

rjurney: @SemyonSinchenko sorry, what about this? I was confused - I see checkpointing happening elsewhere.

```python
cls.spark._jsc.setCheckpointDir(cls.checkpointDir)
```

python/graphframes/tests.py
rjurney: Oh, it was using the freaking `SparkContext` - roger!

python/graphframes/tests.py
rjurney: It is gone!

python/graphframes/tests.py
rjurney: Done.

python/graphframes/tests.py
rjurney: > Can we avoid any usages of the SparkContext at all and rely only on a SparkSession? It allows smoothly test both spark classic and spark connect. You can see how it is done in #506

I don't think so, no. At least I do not know how to accomplish that. My attempt went badly :)

SemyonSinchenko (approved):
LGTM overall! I think we should merge it because in #506 I need to change testing fixtures again and there is no clear way to avoid it. So, let's merge it?

rjurney:
@SemyonSinchenko thanks for the flexibility - merging!

PR #527, opened 2025-02-27
chore: force python code style check in CI

SemyonSinchenko:

- code style (black, flake8, isort) is checking in CI
- all the errors from black, flake8 and isort are fixed

At the moment we have a code style but we are not checking it during the CI


SemyonSinchenko:
@rjurney @bjornjorgensen Hello! This one is quite small, only formatting + CI. Could you take a look when you have time? Thanks!

SemyonSinchenko:
I'm thinking it is a bug in the black (even if we denied it): it won't format comment blocks or break long strings. At the same time, flake8 is still checking line-length for block comments and if we want to have both of them, we should add `noqa: E501` to some of block-comments. And that is exactly what I did in this PR.

python/graphframes/lib/pregel.py
```diff
@@ -16,13 +16,17 @@
...
+if sys.version > "3":
     basestring = str
 
+from pyspark.ml.wrapper import JavaWrapper
 from pyspark.sql import DataFrame, SparkSession
 from pyspark.sql.functions import col
-from pyspark.ml.wrapper import JavaWrapper
+
+if TYPE_CHECKING:
+    from graphframes import GraphFrame
```
bjornjorgensen, line 29: is this right?

python/pyproject.toml
```diff
@@ -30,7 +30,7 @@ pyspark = "^3.4"
 numpy = ">= 1.7"
 
 [tool.poetry.group.dev.dependencies]
-black = "^25.1.0"
+black = "^23.12.1"
```
bjornjorgensen, line 33: why do you downgrade?

python/pyproject.toml
```diff
@@ -50,9 +50,7 @@ build-backend = "poetry.core.masonry.api"
 line-length = 100
 target-version = ["py39"]
 include = ["graphframes"]
-
-[tool.flake8]
```
bjornjorgensen, line 54: apache spark does use I think its 88.. but why do you remove it?

python/poetry.lock
```diff
@@ -976,13 +941,12 @@ test = ["coverage[toml] (>=5.2)", "hypothesis", "pytest (>=6.0)", "pytest-benchm
 
 [[package]]
 name = "pyspark"
-version = "3.5.4"
+version = "3.5.5"
```
bjornjorgensen, line 944: upgrades?

python/poetry.lock
```diff
@@ -909,7 +875,6 @@ version = "1.1.1"
 description = "PPMd compression/decompression library"
 optional = false
 python-versions = ">=3.9"
-groups = ["tutorials"]
```
bjornjorgensen, line 912: why do you remove this group? 
I don't think it should have been in the python dir.. I think it should have been in a doc dir.. but that's another PR..

python/pyproject.toml
SemyonSinchenko: https://github.com/apache/spark/blob/master/dev/pyproject.toml#L30

python/pyproject.toml
SemyonSinchenko: flake8 does not support pyproject.toml; I added tox.ini (copied from [spark](https://github.com/apache/spark/blob/master/dev/tox.ini))

python/poetry.lock
SemyonSinchenko: We are relying on the latest one in CI, I just updated lock-file (by `poetry lock`)

python/poetry.lock
SemyonSinchenko: It is automatically generated by `poetry lock` and is relying on the current structure of `pyproject.toml`

python/graphframes/lib/pregel.py
SemyonSinchenko: Yes. It is needed in the only line and only for type hints. `TYPE_CHECKING` variable is exactly for that case when we need something only during the static analysis, but not in the runtime.

SemyonSinchenko: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

python/poetry.lock
bjornjorgensen: 👍

python/pyproject.toml
bjornjorgensen: ah.. thanks.

PR #536, opened 2025-03-08
Reproduces broken dependabot update of `black` ^24.10.0

rjurney:

Reproduced `black = "^24.10.0"` update in #528 but it shall pass checks


rjurney:
No idea why it fails... @SemyonSinchenko any ideas? Not a high priority but maybe an issue down the line. I edited `pyproject.toml` to indicate `black` `^24.10.0` and I do `poetry update` and it installs... this thing complains...

EDIT: okay when I run the actual command below I get the same output. I don't understand this bug.

```bash
poetry run python -m black --check graphframes
Oh no! 💥 💔 💥 The required version `23.12.1` does not match the running version `24.10.0`!
```

I am running this `poetry`:

```bash
poetry --version
Poetry (version 2.1.1)
```

SemyonSinchenko:
@rjurney so you really want to merge it? I mean, what is the problem with the current version of black, that matches with the version in spark? By the way, the cause of fault is in pyproject toml, in the [tool.black] section.

SemyonSinchenko:
https://github.com/graphframes/graphframes/blob/master/python%2Fpyproject.toml#L54

rjurney:
> @rjurney so you really want to merge it? I mean, what is the problem with the current version of black, that matches with the version in spark? By the way, the cause of fault is in pyproject toml, in the [tool.black] section.

It was a dependabot commit, I don't really care about it. We can leave it, if Spark uses the old version.

Issue #532, opened 2025-03-05
feat: add small-medium sized benchmarks


SemyonSinchenko:
It is very hard to investigate the performance improvements or degradation without a good benchmark, that is integrated to the CI.

**How I see it**
- It should be based on JMH; even if it is a Java tool, it should be an easy task to integrate it to the sbt: there is [plugin](https://github.com/sbt/sbt-jmh) for it;
- It should be possible to run the whole benchmark less that in a 30 minutes, otherwise we may face problems with GitHub limits;
- I don't think that we should run it on each commit, but it is still nice to have and we should at least run it for any commit to the main!

I think we can start from the small-medium sized dataset from [LDBC](https://ldbcouncil.org/benchmarks/graphalytics/). There are some datasets of the size about 100-500 Mb that sounds fine for me.

@SauronShepherd you mentioned in the mailing list that you are interesting in the performance topic. What do you think about it?

cc: @rjurney

SauronShepherd:
Well, I find benchmarks quite interesting, but I'm not a fan of them
because they—like unit tests—can create a false sense of 'everything's
fine' when that might not be the case. They also tend to focus on certain
areas while neglecting others. However, I still think it's a good idea.

Count on me.

rjurney:
@SemyonSinchenko this is why I worked so hard on the `stats.meta` KG build, thinking it could be used to exercise GraphFrames. Ohhh it includes answers for the algorithms on the data? That is badass :)

For reference: https://github.com/graphframes/graphframes/tree/master/python/graphframes/tutorials

rjurney:
This would be great to test out #544 :)

PR #547, opened 2025-03-19
[DO NOT MERGE] feat: LDBC benchmark and tests

SemyonSinchenko:

1. Testing GraphFrames on LDBC graphs + checking the results with a references
2. JMH benchmarks on LDBC graphs

1. We should make GraphFrames be able to pass the LDBC at least with correct results
2. Initial development of performance investigations (as described in #532)

### DIFF
--- a/src/main/scala/org/graphframes/examples/LDBCUtils.scala
+++ b/src/main/scala/org/graphframes/examples/LDBCUtils.scala
@@ -1,0 +1,111 @@
+package org.graphframes.examples
+
+import org.apache.spark.sql.functions.{col, split}
+import org.apache.spark.sql.types.LongType
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.graphframes.GraphFrame
+
+import java.net.URL
+import java.nio.file._
+import java.util.Properties
+import scala.sys.process._
+
+object LDBCUtils {
+  // TODO: This object can be actually a class, parametrized by the id of LDBC graph (kgs, etc.)
+  private val caseName = "kgs" // XS, 65.7Mb
+  private val LDBCUrl = new URL(
+    s"https://datasets.ldbcouncil.org/graphalytics/${caseName}.tar.zst")
+  private val bufferSize = 8192 // 8Kb
+  private val resourcesDir: Path = sys.env.get("LDBC_TEST_ROOT") match {
+    case Some(s) => Paths.get(s).resolve(caseName)
+    case None => throw new RuntimeException("LDBC_TEST_ROOT is not set!")
+  }
+  private val archivePath: Path = resourcesDir.resolve(s"${caseName}.tar.zst")
+  //  Unreachable vertices should be given the value infinity (represented as 9223372036854775807).
+  val UNREACHABLE_ID_VALUE = 9223372036854775807L
+
+  private def checkZSTD(): Unit = {
+    try {
+      s"zstd --version".!
+    } catch {
+      case e: Exception =>
+        throw new RuntimeException(
+          "zstd is not available or not found. Please install zstd and try again.",
+          e)
+    }
+  }
+  def downloadLDBCIfNotExists(): Unit = {
+    if (Files.notExists(resourcesDir) || (Files.list(resourcesDir).count() == 0L)) {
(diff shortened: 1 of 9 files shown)


src/main/scala/org/graphframes/examples/LDBCUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+  def downloadLDBCIfNotExists(): Unit = {
+    if (Files.notExists(resourcesDir) || (Files.list(resourcesDir).count() == 0L)) {
+      println("LDBC data not found. Downloading...")
+      checkZSTD()
+      if (Files.notExists(resourcesDir)) {
+        Files.createDirectory(resourcesDir)
+      }
+      val connection = LDBCUrl.openConnection()
```
rjurney, line 44: We’re doing this manually to avoid a dependency, I take it?

src/main/scala/org/graphframes/examples/LDBCUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+    val vertices = spark.read
+      .text(
+        resourcesDir.resolve(properties.getProperty(s"graph.${caseName}.vertex-file")).toString)
+      .select(col("value").cast(LongType).alias(GraphFrame.ID))
+
+    val edges = spark.read
+      .text(resourcesDir.resolve(properties.getProperty(s"graph.${caseName}.edge-file")).toString)
+      .withColumn("split", split(col("value"), " "))
+      .select(
+        col("split").getItem(0).cast(LongType).alias(GraphFrame.SRC),
```
rjurney, line 85: Same thing here, for CSV?

src/test/scala/org/graphframes/ldbc/LDBCSuiteBFS.scala
```diff
@@ -0,0 +1,55 @@
...
+
+class LDBCSuiteBFS extends SparkFunSuite with GraphFrameTestSparkContext {
+  test("Test LDBC BFS") {
+    assume(sys.env.contains("LDBC_TEST_ROOT"))
+    LDBCUtils.downloadLDBCIfNotExists()
+
+    val enabled = spark.conf.getOption("spark.sql.adaptive.enabled")
+    try {
+      // disable AQE
+      spark.conf.set("spark.sql.adaptive.enabled", value = false)
```
rjurney, line 17: Why disable AQE? I’ve found some algorithms can be painfully slow without it. Avoid bugs?

src/test/scala/org/graphframes/ldbc/LDBCSuiteBFS.scala
```diff
@@ -0,0 +1,55 @@
...
+        .filterNot(_.getAs[Boolean]("matched"))
+        .map(r =>
+          (
+            r.getAs[Long](GraphFrame.ID),
+            r.getAs[Long]("distance"),
+            r.getAs[Long]("expectedDistance")))
+
+      assert(nonMatchedRows.isEmpty)
+    } finally {
+      if (enabled.isDefined) {
```
rjurney, line 48: Don’t we need to test with and without AQE?

rjurney (approved):
lgtm

src/main/scala/org/graphframes/examples/LDBCUtils.scala
SemyonSinchenko: @rjurney For me, the perfect solution would be to grate the repository `graphframes/graphframes-testing`, place all the LDBC data here, add downloading scripts and use it as a git submodule in main graphframes. What do you think about it?

src/test/scala/org/graphframes/ldbc/LDBCSuiteBFS.scala
SemyonSinchenko: This PR at the moment is for testing / sharing / discussions. I was trying to disable the AQE after thread in mailing list

src/test/scala/org/graphframes/ldbc/LDBCSuiteBFS.scala
SemyonSinchenko: Outdated.

PR #563, opened 2025-04-02
Allow setting default ConnectedComponents configurations from spark config

james-willis:
Enable setting the connected components algorithm from spark configurations.

spark configurations are a standard way to control implementation details across a spark runtime. This pattern will help runtime environment owners control how their users use graphframes

### DIFF
--- a/src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
+++ b/src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
@@ -1,0 +1,96 @@
+package org.apache.spark.sql.graphframes
+
+import org.apache.spark.internal.config.ConfigEntry
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.storage.StorageLevel
+
+object GraphFramesConf {
+  private val CONNECTED_COMPONENTS_ALGORITHM =
+    SQLConf
+      .buildConf("spark.graphframes.connectedComponents.algorithm")
+      .doc(""" Sets the connected components algorithm to use (default: "graphframes"). Supported algorithms
+          |   - "graphframes": Uses alternating large star and small star iterations proposed in
+          |     [[http://dx.doi.org/10.1145/2670979.2670997 Connected Components in MapReduce and Beyond]]
+          |     with skewed join optimization.
+          |   - "graphx": Converts the graph to a GraphX graph and then uses the connected components
+          |     implementation in GraphX.
+          | @see org.graphframes.lib.ConnectedComponents.supportedAlgorithms""".stripMargin)
+      .version("0.9.0")
+      .stringConf
+      .createOptional
+
+  private val CONNECTED_COMPONENTS_BROADCAST_THRESHOLD =
+    SQLConf
+      .buildConf("spark.graphframes.connectedComponents.broadcastthreshold")
+      .doc(""" Sets broadcast threshold in propagating component assignments (default: 1000000). If a node
+          | degree is greater than this threshold at some iteration, its component assignment will be
+          | collected and then broadcasted back to propagate the assignment to its neighbors. Otherwise,
+          | the assignment propagation is done by a normal Spark join. This parameter is only used when
+          | the algorithm is set to "graphframes".""".stripMargin)
+      .version("0.9.0")
+      .intConf
+      .createOptional
+
+  private val CONNECTED_COMPONENTS_CHECKPOINT_INTERVAL =
+    SQLConf
+      .buildConf("spark.graphframes.connectedComponents.checkpointinterval")
+      .doc(""" Sets checkpoint interval in terms of number of iterations (default: 2). Checkpointing
+          | regularly helps recover from failures, clean shuffle files, shorten the lineage of the
+          | computation graph, and reduce the complexity of plan optimization. As of Spark 2.0, the
+          | complexity of plan optimization would grow exponentially without checkpointing. Hence,
+          | disabling or setting longer-than-default checkpoint intervals are not recommended. Checkpoint
+          | data is saved under `org.apache.spark.SparkContext.getCheckpointDir` with prefix
+          | "connected-components". If the checkpoint directory is not set, this throws a
+          | `java.io.IOException`. Set a nonpositive value to disable checkpointing. This parameter is
+          | only used when the algorithm is set to "graphframes". Its default value might change in the
+          | future.
+          | @see `org.apache.spark.SparkContext.setCheckpointDir` in Spark API doc""".stripMargin)
+      .version("0.9.0")
+      .intConf
+      .createOptional
+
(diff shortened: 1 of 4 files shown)


SauronShepherd:
latm (= looks asymmetric to me)

Why allow only the `algorithm`? Why not also `checkpointInterval`, `broadcastThreshold`, and `intermediateStorageLevel`?
Why not extend this to all the other classes in GraphFrames?

rjurney:
I like where you're going with this...

james-willis:
> Why allow only the algorithm?

Because I'm lazy and greedy. 

And I'm gauging interest in this kind of approach.

SauronShepherd:
> > Why allow only the algorithm?
> 
> Because I'm lazy and greedy.

Didn't know lazy people actually propose new things ... 😂

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -44,7 +42,9 @@ class ConnectedComponents private[graphframes] (private val graph: GraphFrame)
 
   import org.graphframes.lib.ConnectedComponents._
 
-  private var broadcastThreshold: Int = 1000000
+  private var broadcastThreshold: Int = SparkSession.getActiveSession.get.conf
+    .get("spark.graphframes.connectedComponents.broadcastthreshold", "1000000")
```
SemyonSinchenko, line 46: What do you think about creating an object (as part of the public API) with constants and use them, instead of string-literals? It will allow us to have a slightly more transparent code and at the same time, such an object can be a good starting place for users of graphframes to realize at least what configs can be changed in graphframes? Something like `GraphFramesConfig` or `GraphFramesSessionConfig`?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -71,7 +71,9 @@ class ConnectedComponents private[graphframes] (private val graph: GraphFrame)
    */
   def getBroadcastThreshold: Int = broadcastThreshold
 
-  private var algorithm: String = ALGO_GRAPHFRAMES
+  private var algorithm: String = SparkSession.getActiveSession.get.conf
+    .get("spark.graphframes.connectedComponents.algorithm", ALGO_GRAPHFRAMES)
+    .toLowerCase
```
SemyonSinchenko, line 76: Because it is an internal constant, would it be better to just use lowercase for it instead of having this `toLowerCase` everywhere?

SemyonSinchenko:
@james-willis jfyi, you can use a `pre-commit` hook to be sure that the code is properly formatted: https://github.com/graphframes/graphframes/blob/master/CONTRIBUTING.md#styleguides

Alternatively, just run `./build/sbt root/scalafmtAll`

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
SauronShepherd: Yep, that's a good idea. But not only that ... shouldn't these new configurations be properly documented in the GraphFrames docs? `GraphFramesConfig` sounds good to me.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
SauronShepherd: And what about using the same configuration mechanism as Spark? 

```

  private[spark] val DRIVER_CORES = ConfigBuilder("spark.driver.cores")
    .doc("Number of cores to use for the driver process, only in cluster mode.")
    .version("1.3.0")
    .intConf
    .createWithDefault(1)
```

SauronShepherd:
Hey, and what about the Spark Connect feature in Graphframes? isn't it affected also by this change?

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
james-willis: Sure. We use that pattern in Sedona. I will replicate it here.

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
james-willis: I call `toLowerCase` under the assumption the user might do something funny like 'Graphx' or 'GRAPHX'

src/main/scala/org/graphframes/lib/ConnectedComponents.scala
james-willis: This is a pattern we have used in sedona and I can replace here.

james-willis:
> Hey, and what about the Spark Connect feature in Graphframes? isn't it affected also by this change?

The new interface here is just spark config so I don't think there is any need to change anything about the connect integration but if you think there is some change to be made please point me in the right direction.

SemyonSinchenko:
> Hey, and what about the Spark Connect feature in Graphframes? isn't it affected also by this change?

It isnt, there is no problem, it is just an alternative way of passing configs.

SemyonSinchenko:
@james-willis in the latest changes we introduced a new mixin (WithAlgorithmChoice) to avoid code repeating that was added to ConnectedComponents. So, instead of redefining the default you should probably just call a setter.

james-willis:
Bogged down at work. will get back to this eventually.

src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
```diff
@@ -0,0 +1,93 @@
+package org.apache.spark.sql.graphframes
+
+import org.apache.spark.internal.config.ConfigEntry
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.storage.StorageLevel
+
+object GraphFramesConf {
+  private val CONNECTED_COMPONENTS_ALGORITHM =
+    SQLConf.buildConf("spark.graphframes.connectedComponents.algorithm")
```
james-willis, line 10: only `SQLConf`s are runtime mutable

src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
```diff
@@ -0,0 +1,93 @@
...
+      .doc(""" Sets the connected components algorithm to use (default: "graphframes"). Supported algorithms
+          |   - "graphframes": Uses alternating large star and small star iterations proposed in
+          |     [[http://dx.doi.org/10.1145/2670979.2670997 Connected Components in MapReduce and Beyond]]
+          |     with skewed join optimization.
+          |   - "graphx": Converts the graph to a GraphX graph and then uses the connected components
+          |     implementation in GraphX.
+          | @see org.graphframes.lib.ConnectedComponents.supportedAlgorithms""".stripMargin)
+      .version("0.9.0")
+      .stringConf
+      .createOptional
```
james-willis, line 20: I was thinking we delegate the default value to the trait or the class implementing the trait. Hence set this as optional

SemyonSinchenko (approved):
LGTM! Nice work @james-willis !

PR #603, opened 2025-06-11
feat: Spark 4.0.x support

### LINKED ISSUE #576: [EPIC] Spark 4.0 support
Spark 4.0 was released.

- [x] Run scala tests against spark 4.0
- [x] Run python tests against spark 4.0
- [x] Update spark-connect && graphframes-connect

SemyonSinchenko:
Everything related to the Spark 4.0.x support.

Close #576

### DIFF
--- a/python/dev/build_jar.py
+++ b/python/dev/build_jar.py
@@ -9,7 +9,7 @@ def build(spark_versions: Sequence[str] = ["4.0.0"]):
     for spark_version in spark_versions:
         print("Building GraphFrames JAR...")
         print(f"SPARK_VERSION: {spark_version[:3]}")
-        assert spark_version[:3] in {"4.0",}, "Unsopported spark version!"
+        assert spark_version[:3] in {"4.0",}, "Unsupported spark version!"
         project_root = Path(__file__).parent.parent.parent
         sbt_executable = project_root.joinpath("build").joinpath("sbt").absolute().__str__()
         sbt_build_command = [
--- a/src/main/scala/org/graphframes/lib/ShortestPaths.scala
+++ b/src/main/scala/org/graphframes/lib/ShortestPaths.scala
@@ -102,9 +102,11 @@ private object ShortestPaths extends Logging {
       val longIdToLandmarkColumn = map(longIdToLandmarkFlatten: _*)
       transform_keys(
         col(DISTANCE_ID),
-        (kk: Column, vv: Column) =>
-          map_values(map_filter(longIdToLandmarkColumn, (kkk: Column, _: Column) => kkk === kk))(
-            0))
+        (longId: Column, distance: Column) =>
+          map_values(
+            map_filter(
+              longIdToLandmarkColumn,
+              (mappingId: Column, _: Column) => longId === mappingId))(0))
     }
     val cols = graph.vertices.columns.map(quote).map(col) :+ distanceCol.as(DISTANCE_ID)
     g.vertices.select(cols.toSeq: _*)


src/main/scala/org/graphframes/lib/ShortestPaths.scala
```diff
@@ -90,35 +89,22 @@ class ShortestPaths private[graphframes] (private val graph: GraphFrame)
...
-      val mapToLandmark = udf(func, MapType(idType, IntegerType, false))
-      mapToLandmark(col(DISTANCE_ID))
+      val longIdToLandmarkFlatten: Seq[Column] = longIdToLandmark.flatMap {
+        case (k: Long, v: Any) => Seq(lit(k), lit(v))
+      }.toSeq
+      val longIdToLandmarkColumn = map(longIdToLandmarkFlatten: _*)
+      transform_keys(
+        col(DISTANCE_ID),
+        (kk: Column, vv: Column) =>
+          map_values(map_filter(longIdToLandmarkColumn, (kkk: Column, _: Column) => kkk === kk))(
```
SemyonSinchenko, line 106: I know, naming (`kk`, `kkk`, ...) is terrible :)

src/main/scala/org/graphframes/lib/BFS.scala
```diff
@@ -230,19 +234,36 @@ private object BFS extends Logging with Serializable {
...
+   *   The sequence of valid field names that can appear in the expression.
    * @return
-   *   SQL expression applied to the column fields, such as `myVertex.id = 3`
+   *   A new Column with the expression applied to the column fields, such as `myVertex.age > 30`.
    */
-  private def applyExprToCol(expr: Column, colName: String) = {
-    new Column(expr.expr.transform { case UnresolvedAttribute(nameParts) =>
-      UnresolvedAttribute(colName +: nameParts)
-    })
+  private def applyExprToCol(c: Column, colName: String, fieldNames: Seq[String]): Column = {
```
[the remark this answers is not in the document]
SemyonSinchenko: Don't worry about connect! It will work via Plugin anyway, so we do not need to even care about this. And thanks for the snippet!!!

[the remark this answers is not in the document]
SemyonSinchenko: I just found `org.apache.spark.sql.classic.RichColumn` that provides `expr: Expression`, so maybe there is no need to add a shim...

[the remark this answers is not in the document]
Kimahriman: Well the shim would be for also maintaining support for Spark 3 as well

SemyonSinchenko:
I'm really surprised that it works!

SemyonSinchenko:
@SauronShepherd @james-willis @rjurney Hi! Could you take a look please?

**_Note_**: the target branch is not `main` but `spark-4x`

python/dev/build_jar.py
```diff
@@ -5,11 +5,11 @@
 from pathlib import Path
 
 
-def build(spark_versions: Sequence[str] = ["3.5.5"]):
+def build(spark_versions: Sequence[str] = ["4.0.0"]):
     for spark_version in spark_versions:
         print("Building GraphFrames JAR...")
         print(f"SPARK_VERSION: {spark_version[:3]}")
-        assert spark_version[:3] in {"3.5",}, "Unsopported spark version!"
+        assert spark_version[:3] in {"4.0",}, "Unsopported spark version!"
```
james-willis, line 12: nit: should be Unsupported

src/main/scala/org/graphframes/lib/ShortestPaths.scala
james-willis: probably I should not approve a PR with kkk as a variable name: https://en.wikipedia.org/wiki/Ku_Klux_Klan

james-willis:
I am a little worried that we are not targeting a single branch that supports both spark 3 and 4.

We will probably want to support spark 3 for some time and this might increase the maintainance burden of brining patches and features to both spark 3 and 4. 

In sedona we have directories that have copies of the same code for different versions of spark: https://github.com/apache/sedona/tree/master/spark

not sure if that pattern would succeed here.

SemyonSinchenko:
> I am a little worried that we are not targeting a single branch that supports both spark 3 and 4.
> 
> We will probably want to support spark 3 for some time and this might increase the maintainance burden of brining patches and features to both spark 3 and 4.
> 
> In sedona we have directories that have copies of the same code for different versions of spark: https://github.com/apache/sedona/tree/master/spark
> 
> not sure if that pattern would succeed here.

That is the big question. In GraphFrames there are:
- core
- spark connect
- pyspark classic
- pyspark connect

In the nearest future I'm going to add also property graphs support as a subproject and I hope to have i/o subproject too (for the support of import/export from too GraphAr, Neo4j, Nebula, etc.)

And all of these projects needs to be copied to support both 3.5.x and 4.0.x from the same branch. That is why I'm thinking about relying on branches instead of relying on multiple subprojects for different versions of spark.

I do not think that shims and/or reflection magic will work here too just because from 3.5.x to 4.0.x there are breaking changes not only in accesses and methods...

Overall: I'm a terrible release manager, I do not like CI/CD as well as build systems topic, I'm not very experienced in best practices about supporting multiple versions, etc. So, if someone experienced can draw a diagram or something like this about what would be the best way it will be nice!

src/main/scala/org/graphframes/lib/ShortestPaths.scala
SemyonSinchenko: I did not know... Let me fix it of course!

python/dev/build_jar.py
SemyonSinchenko: Fixed!

Kimahriman:
I have a mostly working version of using shims to support 3.x and 4.x I can try to clean up and make a PR for next week

SauronShepherd:
The goal is to support Spark 4 or actually to take advantage and leverage some of the new features of this new version? I have some ideas in mind that may improve some algorithms a lot (not sure about that, but I think it's worith trying it).

Sorry for being away for so long, but I’m back and quite interested in this ticket. I can start helping out next week.

PR #623, opened 2025-07-03
chore: add useLabelsAsComponents configuration for connected components

### LINKED ISSUE #620: feat: introduce a config to not overwrite component ID from long to the label
#320 introduced a breaking change, but I like the overall idea of that PR.

Tbh, for me it looks like cases when user needs a random LONG ID is the case mostly for library developers, not end-users. To allow developers and libraries to maintain the old behavior I want to introduce a config `spark.graphframes.useLabelsAsComponents` with a default value `true`

1. The new configuration `spark.graphframes.useLabelsAsComponents`
2. The new behavior of CC post-processing

**Component**

- [x] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**
.

SemyonSinchenko:

Introduced a new configuration option `spark.graphframes.useLabelsAsComponents` to allow connected components to use labels as components. Updated the logic in `ConnectedComponents` and added relevant test cases for both enabled and disabled configurations.

Close #620

### DIFF
--- a/src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
+++ b/src/main/scala/org/apache/spark/sql/graphframes/GraphFramesConf.scala
@@ -10,7 +10,7 @@ object GraphFramesConf {
     SQLConf
       .buildConf("spark.graphframes.useLabelsAsComponents")
       .doc(""" Tells the connected components algorithm to use (default: "true") labels as components in the output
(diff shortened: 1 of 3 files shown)


Kimahriman:
LGTM, maybe just mention the components are longs when the config is false in the description?

james-willis (approved):

PR #632, opened 2025-07-15
allow useLabelsAsComponents to be set locally on the ConnectedComponents instance

james-willis:
This change lets useLabelsAsComponents to be set locally rather than globally via a spark config. This lets libraries that consume graphframes to avoid leaky abstraction and enables other usecases where affecting the global state of the cluster is inappropriate. See discussion here: https://github.com/apache/sedona/pull/2098

- Adding the WithUseLabelsAsComponents mixing and add to ConnectedComponents
- change default to false to avoid breaking changes to the ConnectedComponents interface
- 
1. avoid making a breaking change to the return type of this function
2. enable cleaner configuration of the algorithm when it is an implementation detail of a downstream interface


SemyonSinchenko:
Or just add it to the core and I will do all the corresponding updates of PySpark APIs (both classic and connect)

SemyonSinchenko (approved):

PR #634, opened 2025-07-16
feat: set plugin and update pyproject

### LINKED ISSUE #633: feat: add poetry-dynamic-versioning plugin
During the last check of an existing publishing CI I found an issue. `sbt-ci-release` requires (strictly) git tags to have a "v"-prefix, like "v0.9.0" (not just "0.9.0"). But an existing python-publish CI is just [calling](https://github.com/graphframes/graphframes/blob/master/.github/workflows/python-publish.yml#L43) `poetry version ${{ github.ref_name }}` that will set the version to "v0.9.0" (not just "0.9.0"). But "v0.9.0" is not a valid version for the PyPi central and most probably the release will be rejected.

Add https://pypi.org/project/poetry-dynamic-versioning/ as poetry plugin as described in it's documentation. It will carefully extract the version from git tag during the build, so it will allow to simplify the CI as a side effect (the line `poetry version ${{ github.ref_name }}` won't be needed anymore).

**Component**

- [ ] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [x] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**
Highest priority, it blocks the 0.9.0 release.

SemyonSinchenko:
- updates to `pyproject`
- poetry-dynamic-versioning plugin

Close #633 

- Update tags
- Update the structure to the latest poetry (2+)
- Re-generate .lock

P.S. I hope it will work.
P.P.S. We should think about moving to `uv` because it provides much better developer experience. Poetry becomes more and more complex with each version + they are constantly making breaking changes.

### DIFF
--- a/python/graphframes/__init__.py
+++ b/python/graphframes/__init__.py
@@ -1,3 +1,4 @@
 from .graphframe import GraphFrame
 
 __all__ = ["GraphFrame"]
+__version__ = "0.0.0"
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -38,7 +38,7 @@ jobs:
     - name: Install and configure Poetry
       uses: snok/install-poetry@v1
       with:
-        version: 2.1.1
+        version: 2.1.3
         virtualenvs-create: true
         virtualenvs-in-project: false
         installer-parallel: true
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -20,27 +20,27 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up JDK
-        if: startsWith(github.ref, 'refs/tags/')
         uses: actions/setup-java@v4
         with:
           java-version: '11'
           distribution: 'zulu'
 
       - name: Set up Python
-        if: startsWith(github.ref, 'refs/tags/')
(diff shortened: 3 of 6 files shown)


python/graphframes/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .graphframe import GraphFrame
 
 __all__ = ["GraphFrame"]
+__version__ = "0.0.0"
```
james-willis, line 4: can you leave a comment why the version is 0.0.0

james-willis (approved):

python/graphframes/__init__.py
SemyonSinchenko: From the plugin's doc:
> Poetry still requires the tool.poetry.version field to be present in pyproject.toml, but you are encouraged to use version = "0.0.0" as a standard placeholder.
>
> With the minimal configuration above, the plugin will automatically take effect when you run commands such as poetry build. It will update the version in pyproject.toml, then revert the change when the plugin deactivates.
>
> The default configuration will also update any pre-existing __version__ = "0.0.0" and __version_tuple__ = (0, 0, 0) placeholders in some files. You can configure additional substitution patterns/files as needed (see below).

PR #662, opened 2025-07-20
feat: localCheckpoints + docs

### LINKED ISSUE #651: feat: use localCheckpoint in iterative algorithms
Related to #593 

See [comment](https://github.com/graphframes/graphframes/pull/594#issuecomment-3085472410)

**Component**

- [x] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**
.

SemyonSinchenko:
- optional configuration to use `localCheckpoint` instead of `checkpoint` or `parquet`
- updated tests
- new documentation page with GraphFrames configurations
- small updates and fixes in docs

Close #651

P.S. I will be honest, that I used AI-agent to generate `configurations.md` but I checked it manually and it looks OK for me.

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
+++ b/core/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
@@ -207,8 +207,10 @@ object ConnectedComponents extends Logging {
       val logPrefix = s"[CC $runId]"
       logInfo(s"$logPrefix Start connected components with run ID $runId.")
 
+      val shouldUseLocalCheckpoints = GraphFramesConf.getUseLocalCheckpoints
       val shouldCheckpoint = checkpointInterval > 0
-      val checkpointDir: Option[String] = if (shouldCheckpoint) {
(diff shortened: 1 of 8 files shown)


SemyonSinchenko:
I'm thinking about using local checkpoints as defaults starting from 1.0 release to avoid problems with Spark Connect, k8s, etc. where using persistent storage is not always possible or may be hard to configure.

core/src/main/scala/org/graphframes/lib/ConnectedComponents.scala
```diff
@@ -207,8 +207,10 @@ object ConnectedComponents extends Logging {
       val logPrefix = s"[CC $runId]"
       logInfo(s"$logPrefix Start connected components with run ID $runId.")
 
+      val shouldUseLocalCheckpoints = GraphFramesConf.getUseLocalCheckpoints
```
james-willis, line 210: I think we should use mxins that let you set these configs locally with precendence over the global spark configs.

james-willis (approved):

PR #685, opened 2025-09-13
feat: variable length pattern matching

goungoun:
This PR implements variable-length pattern matching. e.g. `(u)-[*3]->(v)`, `(u)-[*1..3]->(v)`
https://s3.amazonaws.com/artifacts.opencypher.org/openCypher9.pdf (page 15)

Users must manually write a long pattern without the feature. After the change, the parser can handle the pattern by creating intermediate vertices and edges. For example `(u)-[*3]->(v)` is converted to `(u)-[]->(_v1);(_v1)-[]->(_v2);(_v2)-[]->(v);` internally. It also supports named edge `(u)-[e*3]->(v)` creating _e1, _e2, _e3. 

The variable length pattern is implemented on top of the fixed-length pattern. It is split into fixed-length pattern. For example `(u)-[*2..5]->(v)` is `(u)-[*2]->(v)` or `(u)-[*3]->(v)` or `(u)-[*4]->(v)` or `(u)-[*5]->(v)`. Then union all the results.

### DIFF
--- a/core/src/main/scala/org/graphframes/pattern/patterns.scala
+++ b/core/src/main/scala/org/graphframes/pattern/patterns.scala
@@ -46,19 +46,20 @@ private[graphframes] object PatternParser extends RegexParsers {
       Negation(e)
     }
   private val fixedLengthPattern: Parser[List[Edge]] =
-    vertex ~ "-" ~ "[" ~ "*" ~ "[0-9]+".r ~ "]" ~ "->" ~ vertex ^^ {
-      case src ~ "-" ~ "[" ~ "*" ~ num ~ "]" ~ "->" ~ dst => {
+    vertex ~ "-" ~ "[" ~ "[a-zA-Z0-9_]*".r ~ "*" ~ "[0-9]+".r ~ "]" ~ "->" ~ vertex ^^ {
+      case src ~ "-" ~ "[" ~ name ~ "*" ~ num ~ "]" ~ "->" ~ dst => {
         val hop: Int = num.toInt
         if (hop == 1) {
-          List(AnonymousEdge(src, dst))
+          List(if (name.isEmpty) AnonymousEdge(src, dst) else NamedEdge(name, src, dst))
         } else if (hop > 1) {
           val midVertices = (1 until hop).map(i => NamedVertex(s"_v$i"))
           val vertices = src +: midVertices :+ dst
           vertices
             .sliding(2)
+            .zipWithIndex
             .map {
-              case Seq(v1, v2) =>
-                AnonymousEdge(v1, v2)
+              case (Seq(v1, v2), i) =>
+                if (name.isEmpty) AnonymousEdge(v1, v2) else NamedEdge(s"_$name${i + 1}", v1, v2)
               case _ => throw new GraphFramesUnreachableException()
             }
             .toList
--- a/core/src/test/scala/org/graphframes/PatternMatchSuite.scala
+++ b/core/src/test/scala/org/graphframes/PatternMatchSuite.scala
@@ -564,6 +564,16 @@ class PatternMatchSuite extends SparkFunSuite with GraphFrameTestSparkContext {
     compareResultToExpected(res, expected)
   }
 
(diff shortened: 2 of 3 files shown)


goungoun:
@SemyonSinchenko Thanks for letting me know. It seems that it failed due to the formatting. I pushed the formatted result by the `sbt scalafmtAll`.

SemyonSinchenko:
> @SemyonSinchenko Thanks for letting me know. It seems that it failed due to the formatting. I pushed the formatted result by the `sbt scalafmtAll`. 

We have quite a strict rules for scalafix... Just add something like case _  => throw new GraphFramesUnreachableException(), should fix it.

SemyonSinchenko (approved):
LGTM! Thanks @goungoun !

goungoun:
Thanks! @SemyonSinchenko.

core/src/main/scala/org/graphframes/pattern/patterns.scala
```diff
@@ -45,8 +45,29 @@ private[graphframes] object PatternParser extends RegexParsers {
     "!" ~ edge ^^ { case _ ~ e =>
       Negation(e)
     }
```
rjurney, line 47: This is such a badly needed feature that... I can't believe it was solved by such few LOC :) Right on!

core/src/test/scala/org/graphframes/PatternMatchSuite.scala
```diff
@@ -547,6 +547,16 @@ class PatternMatchSuite extends SparkFunSuite with GraphFrameTestSparkContext {
     compareResultToExpected(res, Set(Row(0L, 1L, 2L), Row(2L, 0L, 1L), Row(1L, 2L, 0L)))
   }
```
rjurney, line 549: Can we get some more tests? Like a length 5? One just doesn't feel adequate.

rjurney (changes requested):
Please add one more unit test.

core/src/main/scala/org/graphframes/pattern/patterns.scala
goungoun: @rjurney Thank you for reviewing my pull request!

core/src/test/scala/org/graphframes/PatternMatchSuite.scala
goungoun: @rjurney Good point. Unsupported invalid queries too.

goungoun: Increasing the length, it is even more interesting to see how the same node is visited again and again within the cycle.

goungoun:
@rjurney I think `variable length pattern` can be split into `fixed-length pattern`. `(u)-[*2..5]->(v)` is  `(u)-[*2]->(v)` or `(u)-[*3]->(v)` or `(u)-[*4]->(v)` or `(u)-[*5]->(v)`. Then `union` all the results is the expected result. Is that right?

goungoun:
I didn't implement the parser side, still tricky, but it is better to see the code to see the idea.

goungoun:
@rjurney @SemyonSinchenko, There were several commits to cover more cases and edge cases. It is not easy to track changes from the original branch. How can I make it easier for your review? Is it better to open a new PR with a new branch?

SemyonSinchenko:
> How can I make it easier for your review? Is it better to open a new PR with a new branch?

At least if we merge the first PR from you, I won't need anymore to approve each CI run :)

SemyonSinchenko:
@goungoun I would like to mention it again, that we have a `pre-commit` hooks that run the same set of checks like in CI. You can read about it here: https://github.com/graphframes/graphframes/blob/master/CONTRIBUTING.md#styleguides

SemyonSinchenko:
@rjurney what do you think?

rjurney (approved):
lgtm

rjurney:
@goungoun please let us know if you need any support for more contributions! Thanks!

PR #694, opened 2025-09-18
hotfix: java 8 did not have Path.of

SemyonSinchenko:
.

Spark 3.5.x may fail if the Java version is 8. That happens in our build process.
https://github.com/graphframes/graphframes/actions/runs/17837685904/job/50718930629


james-willis (approved):
We should drop Java 8 soon

PR #695, opened 2025-09-19
feat: add validate method

### LINKED ISSUE #49: Consistency checks upon construction

### LINKED ISSUE #283: Provide a method to check graph consistency
We do not always check if vertice IDs in edges exist in vertices due to performance concerns. It would be nice to add those checks and provide a utility method to add those additional vertices to edges to fix graph consistency, or drop those edges. This could be done in two PRs: 1) check consistency, 2) fix consistency.

cc: @felixcheung @jkbradley @MrBago @bohanjason

### LINKED ISSUE #300: Option to prevent deduplicating of already deduplicated vertices or to return already deduplicated ones
From GraphX jobs, when we have to prepare vertices and edges manually, deduplication of vertices is also performed manually.

So GrapFrames can be fed with already deduplicated vertices, but currently it does not know about it and will deduplicate the vertices one more time.
In case of large amount of vertices (e.g. billions) the time of the second deduplication can be significant. 

Moreover the user usually have to deduplicate the vertices in advance.
Consider the following two examples

**Without deduplication**
```
import org.graphframes.GraphFrame

sc.setCheckpointDir("hdfs:///tmp/graphframes-cc")

val v = spark
  .createDataFrame(Seq("1", "2", "3", "4", "5", "5", "4", "3", "2", "1")
  .map(Tuple1(_)))
  .select($"_1".as("id"))
val e = spark
  .createDataFrame(Seq("1"->"2","2"->"3","4"->"5","5"->"4"))
  .select($"_1".as("src"),$"_2".as("dst"))
val g = GraphFrame(v, e)

val cc = g.connectedComponents.run()

cc.orderBy($"id", $"component").show(false)
+---+------------+                                                              
|id |component   |
+---+------------+
|1  |154618822656|
|1  |154618822656|
|2  |154618822656|
|2  |154618822656|
|3  |154618822656|
|3  |154618822656|
|4  |420906795008|
|4  |420906795008|
|5  |420906795008|
|5  |420906795008|
+---+------------+
```
Please note that every vertex occur twice in the output

**With deduplication**
```
import org.graphframes.GraphFrame

sc.setCheckpointDir("hdfs:///tmp/graphframes-cc")

val v = spark
  .createDataFrame(Seq("1", "2", "3", "4", "5", "5", "4", "3", "2", "1")
  .map(Tuple1(_)))
  .select($"_1".as("id"))
  .distinct()
val e = spark
  .createDataFrame(Seq("1"->"2","2"->"3","4"->"5","5"->"4"))
  .select($"_1".as("src"),$"_2".as("dst"))
val g = GraphFrame(v, e)

val cc = g.connectedComponents.run()

cc.orderBy($"id", $"component").show(false)
+---+------------+                                                              
|id |component   |
+---+------------+
|1  |154618822656|
|2  |154618822656|
|3  |154618822656|
|4  |420906795008|
|5  |420906795008|
+---+------------+
```

So it seems that there are two options 

- In case of GraphFrames will always deduplicate vertices during their indexing internally, it seems that it makes sense to return already deduplicated vertices.
- In case of vertices may already be deduplicated it makes sence to prevent their additional deduplication during indexing by providing a corresponding options to the algorithms.

### LINKED ISSUE #488: Validate edge `src` and `dst` field appear in vertices `id` field upon GraphFrame instantiation
You can create `GraphFrames` where there are edges that point at `src` and `dst` `ids` that do not exist in the `id` field of the vertices. I want to improve instantiation by complaining and dying if an INNER join of the edges against the nodes for both `src` and `dst` doesn't produce the same number of edges from a raw edge count.

Something like this:

```python
# Sanity test that all edges have valid ids
edge_count = g.edges.count()
valid_edge_count = (
    g.edges.join(g.vertices, on=g.edges.src == g.vertices.id)
    .select("src", "dst", "relationship")
    .join(g.vertices, on=g.edges.dst == g.vertices.id)
    .count()
)

# Just up and die if we have edges that point to non-existent nodes
assert (
    edge_count == valid_edge_count
), f"Edge count {edge_count} != valid edge count {valid_edge_count}"
print(f"Edge count: {edge_count:,} == Valid edge count: {valid_edge_count:,}")
```

SemyonSinchenko:

From now users should care about graph consistency. We do not need to call distinct and other things. Also, updated indexedVertices / indexedEdges by removing the skewedJoin calls.

Close #300 
Close #488
Close #283
Close #49

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -21,6 +21,7 @@ jobs:
       JVM_OPTS:  -Xms2048M -Xmx2048M -Xss6M -XX:ReservedCodeCacheSize=256M -Dfile.encoding=UTF-8
       SPARK_VERSION: ${{ matrix.spark-version }}
       SCALA_VERSION: ${{ matrix.scala-version }}
+      PIP_REQUESTS_TIMEOUT: 100
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-java@v4
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    env:
+      PIP_REQUESTS_TIMEOUT: 100
     environment:
       name: pypi
       url: https://pypi.org/p/graphframes-py


core/src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -550,6 +551,72 @@ class GraphFrame private (
...
+        s"Graph contains (${verticesCount - countDistinctVertices}) duplicate vertices.")
+    }
+    if (checkVertices) {
+      val verticesSetFromEdges = edges
+        .select(col(SRC).alias(ID))
+        .union(edges.select(col(DST).alias(ID)))
+        .distinct()
+        .persist(intermediateStorageLevel)
+      val countVerticesFromEdges = verticesSetFromEdges.count()
+      if (countVerticesFromEdges != countDistinctVertices) {
```
james-willis, line 601: would this throw if there are vertices that are not referenced in the edge set?

SemyonSinchenko: Fixed

rjurney (approved):
lgtm

PR #698, opened 2025-09-19
feat: performance and API sync

### LINKED ISSUE #219: Make edgeStorageLevel and vertexStorageLevel configurable

### LINKED ISSUE #667: feat: replace rdd.cache() by checkpoint / localCheckpoint in AggregateMessages

This code can be (and should be) replaced by checkpoint. It should be handled in the same way like in other places: if config for using `localCheckpoint` is specified use it, otherwise use `checkpoint`

```scala
  /**
   * Create a new cached copy of a DataFrame. For iterative DataFrame-based algorithms.
   *
   * WARNING: This is NOT the same as `DataFrame.cache()`. The original DataFrame will NOT be
   * cached.
   *
   * This is a workaround for SPARK-13346, which makes it difficult to use DataFrames in iterative
   * algorithms. This workaround converts the DataFrame to an RDD, caches the RDD, and creates a
   * new DataFrame. This is important for avoiding the creation of extremely complex DataFrame
   * query plans when using DataFrames in iterative algorithms.
   */
  def getCachedDataFrame(df: DataFrame): DataFrame = {
    val rdd = df.rdd.cache()
    // rdd.count()
    df.sparkSession.createDataFrame(rdd, df.schema)
  }
```

.

**Component**

- [x] Scala Core Internal
- [x] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**
Performance + code unification

### LINKED ISSUE #696: feat: explicitly select only required columns in all Pregel-based impls
After I added persist to the Pregel it give us a big performance boost. Unfortunately, there is a risk of slow down in case the input Graph contains too much attributes because we are not filtering them explicitly.

Explicitly select only required columns before persisting.

**Component**

- [x] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**

SemyonSinchenko:
- intermediateStorageLevel everywhere
- remove rdd.cache from AM
- selecting required columns in Pregel

Close #696 
Close #667 
Close #219

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/AggregateMessages.scala
+++ b/core/src/main/scala/org/graphframes/lib/AggregateMessages.scala
@@ -23,6 +23,7 @@ import org.apache.spark.sql.functions.col
 import org.apache.spark.sql.functions.expr
 import org.graphframes.GraphFrame
 import org.graphframes.Logging
+import org.graphframes.WithIntermediateStorageLevel
 
 /**
  * This is a primitive for implementing graph algorithms. This method aggregates messages from the
@@ -59,9 +60,13 @@ import org.graphframes.Logging
  */
 class AggregateMessages private[graphframes] (private val g: GraphFrame)
     extends Arguments
-    with Serializable {
+    with Serializable
+    with WithIntermediateStorageLevel
+    with Logging {
 
-  import org.graphframes.GraphFrame.{DST, ID, SRC}
+  import org.graphframes.GraphFrame.DST
+  import org.graphframes.GraphFrame.ID
+  import org.graphframes.GraphFrame.SRC
 
   private var msgToSrc: Option[Column] = None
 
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/graphframes/lib/AggregateMessages.scala
```diff
@@ -59,9 +60,13 @@ import org.graphframes.Logging
  */
 class AggregateMessages private[graphframes] (private val g: GraphFrame)
     extends Arguments
-    with Serializable {
+    with Serializable
+    with WithIntermediateStorageLevel
+    with Logging {
 
-  import org.graphframes.GraphFrame.{DST, ID, SRC}
```
rjurney, line 64: Just curious why this was made into three imports? Strange change.

rjurney (approved):
lgtm

core/src/main/scala/org/graphframes/lib/AggregateMessages.scala
SemyonSinchenko: It is how scalafmt is configured :)

Issue #502, opened 2025-02-07
Add bidirectional edges to network motif finding


rjurney:
The only type of edge atm that is valid is `(a)-[e1]->(b)` and I would like to add support for `(a)<-[e1]->(b)` in [GraphFrame.find()](https://graphframes.github.io/graphframes/docs/_site/api/python/graphframes.html#graphframes.GraphFrame.find). This can include support for backwards facing edges: `(a)<-[e1]-(b)`.

SauronShepherd:
I'm taking this one!

rjurney:
Awesome, this is a badly needed one - it confuses users I'm sure, who are expecting Cypher paths.

SauronShepherd:
I'm adding unit test to the `GraphFrames.find` method. I'm having some problems now with bidirectional motifs, but ... I'm working on it.

```scala
  test("find using different motifs") {
    val patterns = Seq(
      // Find triangles (three nodes forming a cycle):
      //  Finds all cases where three nodes are mutually connected
      "(a)-[]->(b); (b)-[]->(c); (c)-[]->(a)" -> "[[f,m],[f,m],[f,m]]",

      // Find reciprocal (bi-directional) relationships:
      //  Finds all pairs of nodes that have two-way relationships
      "(a)-[]->(b); (b)-[]->(a)" -> "[[b,f],[c,m]],[[c,m],[b,f]],[[f,m],[f,m]]",
      "(a)<-[]->(b)" -> "[[b,f],[c,m]],[[c,m],[b,f]],[[f,m],[f,m]]"
    )

    for((pattern, expected) <- patterns) {
      val result = findGraph.find(pattern).collect().mkString(",")
      assert(result === expected)
    }
  }
```

I've also totally refactorized the unit tests from `PatternSuite` to avoid unneeded verbosity. I'm having some problems now with bidirectional motifs, but ... I'm working on fixing it.

```scala
  test("find named elements in order") {
    val parses: Seq[(String, Seq[String])] = Seq(
      "(u)-[]->(v); (v)-[]->(w); !(u)-[]->(w)" -> Seq("u", "v", "w"),
      "(u)<-[]-(v); (v)<-[]-(w); !(u)<-[]-(w)" -> Seq("v", "u", "w"),
      "(u)<-[]->(v); (v)<-[]->(w); !(u)<-[]->(w)" -> Seq("u", "v", "w"),

[...]

PR #717, opened 2025-10-01
feat: add bidirectional motif pattern 

goungoun:
This PR implements a bidirectional motif pattern as a patterns that exist in both directions.
image
```
g.find("(u)<-[]->(v)").select("u.id", "v.id")

+---+---+
| id| id|
+---+---+
|  0|  1|
|  1|  0|
+---+---+
```

### How does it work?
It rewrites the incoming pattern string before calling the `Pattern.parse` method.
For the bidirectional pattern, `(u)<-[]->(v)` is converted to `(u)-[]->(v);(v)-[]->(u)`
It also supports the reversed pattern `(u)<-[]-(v)`. It converts to the pattern `(v)-[]->(u)`

The feature is based on issue #502. The opinion is that it confuses users without the feature.


SemyonSinchenko (approved):
~LGTM

Issue #713, opened 2025-09-29
feat: redesign python API


SemyonSinchenko:
**Is your feature request related to a problem? Please describe.**
It is really hard to maintain due to two implementations.

**Describe the solution you would like**
- All the logic is in the top-level `graphframes.py` instead of being duplicated in connect / classic
- Naming convention should follow Python, not Scala (snake_case)
- Pregel and AM should be top-level classes without details of the implementation
- Connect / Classic should not contain logic, docs, etc., but should be just simple functions classes to proxy the py4j or protobuf

**Component**

- [ ] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [x] PySpark Classic
- [x] PySpark Connect

**Additional context**
**BREAKING CHANGES**

I see it as a part of the planning 1.0 release.

**Are you planning on creating a PR?**

- [x] I'm willing to make a pull-request

PR #742, opened 2025-10-21
Add Graphframe Type Degree Methods

### LINKED ISSUE #519: feat: create `GraphFrame.typeDegree()` to provide degree by `relationship` type
A call to `GraphFrame.typeDegree()`, `.typeInDegree()` or `.typeOutDegree()` should return a `pyspark.sql.Types.Struct` with a field for each edge type defined in the `GraphFrame`'s `relationships`. So for example:

joelrobin18:
Implementation of Type Degree methods such as typeDegree, typeInDegree, TypeOutDegree

Fix #519

### DIFF
--- a/core/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/core/src/main/scala/org/graphframes/GraphFrame.scala
@@ -318,23 +318,24 @@ class GraphFrame private (
    * @param edgeTypeCol
    *   Name of the column in edges DataFrame that contains edge types
    * @param edgeTypes
-   *   Optional sequence of edge type values. If provided, avoids a collect() operation. If None,
-   *   edge types will be discovered automatically.
+   *   Optional sequence of edge type values. If None, edge types will be discovered
+   *   automatically.
    * @group degree
    */
   def typeOutDegree(edgeTypeCol: String, edgeTypes: Option[Seq[Any]] = None): DataFrame = {
-    import org.apache.spark.sql.functions.coalesce
-
     val pivotDF = edgeTypes match {
       case Some(types) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/graphframes/GraphFrame.scala
```diff
@@ -310,6 +310,83 @@ class GraphFrame private (
...
+  /**
+   * The out-degree of each vertex per edge type, returned as a DataFrame with two columns:
+   *   - [[GraphFrame.ID]] the ID of the vertex
+   *   - "outDegrees" a struct with a field for each edge type, storing the out-degree count
+   *
+   * @param edgeTypeCol
+   *   Name of the column in edges DataFrame that contains edge types
+   * @group degree
+   */
+  def typeOutDegree(edgeTypeCol: String): DataFrame = {
```
SemyonSinchenko: Let's allow users to pass a known list of values of the `edgeTypeCol`? It will benefit both pivot and case-when approach.

python/graphframes/graphframe.py
```diff
@@ -225,6 +225,84 @@ def degrees(self) -> DataFrame:
             .agg(F.count("*").alias("degree"))
         )
 
+    def typeOutDegree(self, edgeTypeCol: str) -> DataFrame:
```
SemyonSinchenko: I'm trying to slowly move python client to a "pythonic" style (#713), so, can I ask to rename it to `type_out_degree(self, edge_type_col)` (and other methods too)?

SemyonSinchenko (approved):

SemyonSinchenko:
Thanks for the contribution @joelrobin18 !

PR #764, opened 2025-12-29
add more LDBC benchmark datasets

james-willis:
Adding some more LDBC graph500 benchmark datasets to make the utils more useful

I'm testing regressions in connected components and need larger datasets

### DIFF
--- a/core/src/main/scala/org/graphframes/examples/LDBCUtils.scala
+++ b/core/src/main/scala/org/graphframes/examples/LDBCUtils.scala
@@ -18,7 +18,14 @@ object LDBCUtils {
   val TEST_WCC_UNDIRECTED = "test-wcc-undirected"
   val KGS = "kgs"
   val GRAPH500_22 = "graph500-22"
+  val GRAPH500_23 = "graph500-23"
   val GRAPH500_24 = "graph500-24"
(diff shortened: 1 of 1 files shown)


SemyonSinchenko (approved):
My only concern is you may face some problems with downloading and unpacking 10th of gigabytes of tar zst archives from inside of JVM.

james-willis:
> My only concern is you may face some problems with downloading and unpacking 10th of gigabytes of tar zst archives from inside of JVM.

Yeah but that is the problem of the user to allocate more resources I suppose.

rjurney:
Is it possible to stream to disk? Then the cluster might not need all the RAM to download? Not sure I understand the issue, just popping my head up from a Lyft :)

james-willis:
Sem's original code does stream the compressed file to disk:

```
      val connection = ldbcURL(name).openConnection()
      val inputStream = connection.getInputStream
      val outputStream = Files.newOutputStream(archivePath)
      val buffer = new Array[Byte](bufferSize)
      var bytesRead = 0
      while ({ bytesRead = inputStream.read(buffer); bytesRead } != -1) {
        outputStream.write(buffer, 0, bytesRead)
      }
      inputStream.close()
      outputStream.close()
```

PR #766, opened 2026-01-02
Fix: Throw IllegalArgumentException for null IDs/Src/Dst in toGraphX (Fixes #765)

### LINKED ISSUE #765: bug: it is possible to reach GraphFramesUnreachableException
**Describe the bug**
If Graph has Long-like ID type and at the same time contains NULLs for src, dst or id the `GraphFramesUnreachableExcpetion` will be raised from [this method](https://github.com/graphframes/graphframes/blob/main/core/src/main/scala/org/graphframes/GraphFrame.scala#L219)

**To Reproduce**

1. ...
2. ...
3. ...

**Expected behavior**

NULLs are not supported of course and it is a case of malformed graph. But exception should be meaningful.

**System [please complete the following information]:**

-   OS: e.g. [Ubuntu 18.04]
-   Python Version (if applied): [e.g. Python 3.8]
-   Spark / PySpark version: [e.g. PySpark 3.5.1]
-   GraphFrames version: [e.g. graphframes-0.9.0]

**Component**

- [x] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**

JeganVG:
…(for #765)

1) Updated the pattern matching logic in toGraphX (for both vertices and edges) to explicitly catch null values.
2) Replaced the generic GraphFramesUnreachableException with a meaningful IllegalArgumentException when null IDs are encountered.

Added Two TCs :

1) toGraphX should throw IllegalArgumentException for null IDs
2) toGraphX should throw IllegalArgumentException for null Edge Src/Dst

Fixes #765
Currently, if a GraphFrame contains null values in the ID (or src/dst) columns and toGraphX is called, the pattern matching falls, throwing a GraphFramesUnreachableException without proper exception.

### DIFF
--- a/core/src/main/scala/org/graphframes/GraphFrame.scala
+++ b/core/src/main/scala/org/graphframes/GraphFrame.scala
@@ -207,25 +207,11 @@ class GraphFrame private (
    */
   def toGraphX: Graph[Row, Row] = {
     if (hasIntegralIdType) {
-      // 1. Null IDs
-      val badVertices = vertices.filter(col(ID).isNull).limit(1).count()
-
-      if (badVertices > 0) {
-        throw new IllegalArgumentException(
-          s"Vertex ID cannot be null. Found null in column '$ID'.")
-      }
-
-      // 2. Null Src/Dst
-      val badEdges = edges.filter(col(SRC).isNull || col(DST).isNull).limit(1).count()
-
-      if (badEdges > 0) {
-        throw new IllegalArgumentException(s"Edge '$SRC' and '$DST' cannot be null.")
-      }
-    }
-
-    if (hasIntegralIdType) {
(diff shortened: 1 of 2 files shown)


SemyonSinchenko:
We have both scalafmt and scalafix. If you do not want to use pre-commit, just call `sbt scalafixAll`

JeganVG:
Hi @SemyonSinchenko,

My bad, I ran the comment and I didn't noticed the error after formatting was done, scalafix and scalafmt passed, I have rebuilt the PR, could you check ?

JeganVG:
@SemyonSinchenko ,

updated the logic, verified test cases and precommit checks locally, could you check ? ;(

PR #769, opened 2026-01-08
Make tutorials work when installing package from PyPi, bump versions of stuff to latest

rjurney:


python/pyproject.toml
```diff
@@ -60,22 +72,19 @@ enable = true
...
-flake8 = "^7.1.1"
-isort = "^6.0.0"
+black = "^25.12.0"
+flake8 = "^7.3.0"
+isort = "^7.0.0"
 pyspark = { version = ">=3.5, <4.1", extras = ["connect"] }
 grpcio = "<=1.67.1"
-pytest = "^8.3.4"
+pytest = "^9.0.2"
+pre-commit = "^4.5.1"
```
rjurney: `pre-commit` wasn't even a dev dependency.

python/pyproject.toml
```diff
@@ -60,22 +72,19 @@ enable = true
...
 
 [tool.poetry.group.tutorials.dependencies]
```
rjurney: Don't want something like `py7zr` installed all the time - just if you want to run the tutorials.

python/pyproject.toml
```diff
@@ -86,10 +95,10 @@ graphframes = "graphframes.console:main"
 
 [tool.black]
 line-length = 100
-target-version = ["py39"]
-include = ["graphframes"]
-
-required-version = "23.12.1"
+target-version = ["py310", "py311", "py312"]
+include = "\\.py$"
+extend-exclude = "\\.parquet$"
```
rjurney: `black` kept trying to format Parquet files...

python/pyproject.toml
```diff
@@ -36,6 +36,18 @@ classifiers = [
...
 dynamic = ["version"]
 requires-python = ">=3.10"
+dependencies = [
+    "numpy>=1.7",
+    "pyspark>=3.5,<4.1",
+    "typing_extensions>=4.0.0",
+]
+
+[project.optional-dependencies]
+tutorials = [
```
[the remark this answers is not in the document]
SemyonSinchenko: Cause they are optional and for tutorials only, I don't care do they have versions or not. Overall, a good approach would be to pin major versions only (like >=1.0,<2.0)

python/graphframes/tutorials/download.py
```diff
@@ -3,18 +3,22 @@
...
 
 import os
+from pathlib import Path
 
 import click
 import py7zr
 import requests  # type: ignore
 
+# Default data directory is relative to this module's location
+DEFAULT_DATA_DIR = str(Path(__file__).parent / "data")
```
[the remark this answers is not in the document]
SemyonSinchenko: Let's better add a cli arg with a folder? Otherwise, for example, if you want to test two different versions of gf, you will download it twice, etc.

PR #789, opened 2026-01-29
feat: add approximate triangle counting algorithm using DataSketches

### LINKED ISSUE #786: feat: approximate triangles count API
The current triangle count is based on the intersection of neighbors sets. It cannot be scaled to power-law graphs. We should have an approximate version of it.

The solution based on intersection of theta-scetches. Should fail if spark version is less than 4.1

**Component**

- [x] Scala Core Internal
- [x] Scala API
- [x] Spark Connect Plugin
- [ ] Infrastructure
- [x] PySpark Classic
- [x] PySpark Connect

**Additional context**

SemyonSinchenko:
This introduces a new "approx" algorithm for triangle counting that leverages Apache Spark's DataSketches (Theta sketches) integration (available since Spark 4.1). This provides a scalable alternative to the existing "exact" implementation, which often encounters Out-of-Memory (OOM) issues on power-law graphs with high-degree nodes.

Changes include:
- Core: Added `approximateRun` to `TriangleCount.scala` using `theta_sketch_agg` and `theta_intersection`.
- Core: Added `GraphFramesRequireSpark` exception for feature gating.
- Connect: Updated protobuf definition and Scala/Python logic to support `algorithm` and `lg_nom_entries` parameters.
- Python: Updated `triangleCount` API in both classic and connect implementations to support the new parameters.
- Docs: Expanded the traversal guide with a detailed comparison between exact and approximate implementations and usage guidelines.
- Tests: Added Scala and Python test suites for the approximate algorithm.

Closes #786

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/TriangleCount.scala
+++ b/core/src/main/scala/org/graphframes/lib/TriangleCount.scala
@@ -190,7 +190,6 @@ private object TriangleCount extends Logging {
         floor(
           (coalesce(col("src_triplets"), lit(0)) + coalesce(col("dst_triplets"), lit(0))) / lit(
             2)))
-      .select(col(ID), col(COUNT_ID))
 
     result.persist(intermediateStorageLevel)
     result.count()


core/src/main/scala/org/graphframes/lib/TriangleCount.scala
```diff
@@ -21,30 +21,58 @@ import org.apache.spark.sql.DataFrame
...
+  private var algorithm: String = "exact"
+  private val supportedAlgorithms: Set[String] = Set("exact", "approx")
+  private var lgNomEntries: Int = 12
+
+  /**
+   * Sets the log2 of the nominal entries for the Theta sketch (only for "approx" algorithm).
+   * Default is 12 (4096 entries).
+   */
+  def setLgNomEntries(value: Int): this.type = {
+    lgNomEntries = value
```
james-willis: can we add a require to assert this is between 4 and 26 as the pyspark function requires?

python/graphframes/classic/graphframe.py
```diff
@@ -323,9 +323,13 @@ def svdPlusPlus(
         v = DataFrame(jdf, self._spark)
         return (v, loss)
 
-    def triangleCount(self, storage_level: StorageLevel) -> DataFrame:
+    def triangleCount(
+        self, storage_level: StorageLevel, algorithm: str, log_nom_entries: int
```
james-willis: should the new parameters to optional to maintain backwards compat?

core/src/main/scala/org/graphframes/lib/TriangleCount.scala
```diff
@@ -65,6 +93,65 @@ private object TriangleCount extends Logging {
...
+            .when(col("src_triplets").isNull, col("dst_triplets"))
+            .when(col("dst_triplets").isNull, col("src_triplets"))
+            .otherwise(col("src_triplets") + col("dst_triplets")) / lit(2)))
+
+    result.persist(intermediateStorageLevel)
+    result.count()
+    verticesWithNeighbors.unpersist()
+    triangles.unpersist()
+    resultIsPersistent()
+    result
```
james-willis: do we need to drop `src_triplets` and `dst_triplets`?

james-willis (changes requested):

python/graphframes/classic/graphframe.py
SemyonSinchenko: Done

james-willis (approved):

PR #825, opened 2026-04-07
feat: add NeighborhoodAwareCDLP community detection algorithm

### LINKED ISSUE #301: semi-supervised LPA
Hi,

Is it possible that nodes be assigned preliminary labels? This can help to find communities where we hand-pick some initial communities.

### LINKED ISSUE #456: Pre-determined settings for Label Propagation Algorithm of Scala version
First of all thank you for this great work. I have got a few questions of Scala LPA algorithm: 
1. Can I set initial labels for each node to accelerate the training?
2. Can I set edge weight?
I would appreciate it if u could help me with my questions. Thank you!

SemyonSinchenko:
- Introduce NeighborhoodAwareCDLP: a neighborhood-aware variant of label propagation that weights incoming votes by a combination of direct-link strength (a) and neighborhood overlap (c * commonNeighbors).
- Add implementation at core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala with:
  - approximate common-neighbor estimation using theta sketches,
  - parameters for a, c, initial label column, and sketch size,
  - Pregel-based propagation and integration with GraphFrame options.
- Expose API on GraphFrame as structureAwareLabelPropagation.
- Add comprehensive unit tests at core/src/test/scala/org/graphframes/lib/NeighborhoodAwareCDLPSuite.scala covering basic propagation, parameter sensitivity, directed/undirected behavior, isolated vertices, and disconnected components.
- Bump default Spark version from 3.5.7 to 3.5.8 in build.sbt.
- Note: the theta-sketch based overlap estimation requires Spark >= 4.1; the implementation checks the Spark version and fails fast on older versions.

The current CDLP is very "basic" but optimized well for it's own problem. I do not want to break it. The new implementation is mostly based on the https://arxiv.org/pdf/1105.3264 with my won adjustments.

image

(on the picture c=0 is a classical CDLP)

Close #791 
Close #301 
Close #456 (partially?)

### Python?

After I get an approve on the core I will add python.


core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
```diff
@@ -0,0 +1,280 @@
...
+      throw new GraphFramesSparkVersionException("4.1.0")
+    }
+
+    val edges = if (isDirected) {
+      graph.edges.select(SRC, DST)
+    } else {
+      graph.edges
+        .select(col(DST).alias(SRC), col(SRC).alias(DST))
+        .union(graph.edges.select(SRC, DST))
+        .distinct()
```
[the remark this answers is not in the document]
SemyonSinchenko: Parallel edges will break the theta sketches.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+   *     edge has equal unit influence.
+   *   - `a = 0` and `c = 1` removes direct-link baseline influence and relies fully on common
+   *     neighbors (pure structure propagation).
+   *
+   * Default: `1.0`.
+   */
+  def setA(value: Double): this.type = {
+    require(value >= 0.0 && value <= 1.0, "a must be in [0,1]")
+    a = value
+    this
```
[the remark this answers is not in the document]
SemyonSinchenko: Not relevant. Reference paper uses 1.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * This value controls Theta sketch capacity (`theta_sketch_agg`) used to estimate overlap
+   * between neighborhood sets. It tunes the quality/resource tradeoff:
+   *   - Larger values improve approximation quality/stability of common-neighbor estimates.
+   *   - Smaller values reduce memory usage (and often compute overhead) but increase
+   *     approximation error.
+   *
+   * Valid range is `[4, 24]`. Default: `12`.
+   */
+  def setLgNomEntries(value: Int): this.type = {
```
james-willis: could become a mixin.

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
```diff
@@ -0,0 +1,280 @@
...
+
+  def run(): DataFrame = {
+    // Validate parameters
+    val maxIterChecked = check(maxIter, "maxIter")
+
+    // Sketch-based features require Spark >= 4.1
+    val sparkVersion = graph.vertices.sparkSession.version
+    if (sparkVersion.substring(0, 3) < "4.1") {
+      throw new GraphFramesSparkVersionException("4.1.0")
+    }
```
james-willis, line 151: claude always gets upset when you use strings for version comparisons. should we refactor all cases of these version checks into a util method and do correct semantic version compares?

core/src/test/scala/org/graphframes/TestUtils.scala
```diff
@@ -21,6 +21,12 @@ object TestUtils {
     }
   }
 
+  /** Return true if the major and minor versions are greater or eq to constraints */
+  def requireSparkVersionGT(major: Int, minor: Int, sparkVersion: String): Boolean = {
+    val (gotMajor, gotMinor) = TestUtils.majorMinorVersion(sparkVersion)
+    (gotMajor >= major) && (gotMinor >= minor)
```
james-willis:
```suggestion
    gotMajor > major || (gotMajor == major && gotMinor >= minor)
```

also change the name to `requireSparkVersionGE`

core/src/test/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
```diff
@@ -0,0 +1,303 @@
+package org.graphframes.lib
```
james-willis, line 18: apache header? do we care about these?

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+   * Key regimes:
+   *   - `c = 0` (with `a = 1`) behaves like classical unweighted label propagation: each incoming
+   *     edge contributes equal vote mass, independent of overlap.
+   *   - `c = 1` and `a = 0` gives pure structure propagation: vote mass depends only on common
+   *     neighbors, so direct links with no shared-neighborhood support contribute zero mass.
+   *
+   * Default: `0.5`.
+   */
+  def setC(value: Double): this.type = {
+    require(value >= 0.0 && value <= 1.0, "c must be in [0,1]")
```
james-willis: what is the important of keeping these between 0 and 1?

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * Key regimes:
+   *   - `c = 0` (with `a = 1`) behaves like classical unweighted label propagation: each incoming
+   *     edge contributes equal vote mass, independent of overlap.
+   *   - `c = 1` and `a = 0` gives pure structure propagation: vote mass depends only on common
+   *     neighbors, so direct links with no shared-neighborhood support contribute zero mass.
+   *
+   * Default: `0.5`.
+   */
+  def setC(value: Double): this.type = {
```
james-willis: should c and a have more descriptive names? could be helpful for users, most of them wont be familiar with the paper

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+    } else {
+      graph.edges
+        .select(col(DST).alias(SRC), col(SRC).alias(DST))
+        .union(graph.edges.select(SRC, DST))
+        .distinct()
+    }
+
+    // Compute approximate common neighbor counts on edges and materialize.
+    val enrichedEdges =
+      computeEdgeApproxCommonNeighbors(edges, lgNomEntries, c, a)
```
james-willis: should we make sure `!(a == 0 && c == 0)`?

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+    extends Arguments
+    with WithCheckpointInterval
+    with WithMaxIter
+    with WithLocalCheckpoints
+    with WithIntermediateStorageLevel
+    with WithDirection
+    with Logging {
+
+  private var c: Double = 0.5
+  private var a: Double = 1.0
```
james-willis: what does the parameter a really matter? isnt the only thing that matters the ratio of a to c? which can be managed with various values of c.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+class NeighborhoodAwareCDLP private[graphframes] (private val graph: GraphFrame)
+    extends Arguments
+    with WithCheckpointInterval
+    with WithMaxIter
+    with WithLocalCheckpoints
+    with WithIntermediateStorageLevel
+    with WithDirection
+    with Logging {
+
+  private var c: Double = 0.5
```
james-willis: why 0.5?

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+    require(value >= 0.0 && value <= 1.0, "c must be in [0,1]")
+    c = value
+    this
+  }
+
+  /**
+   * Sets weight for the direct-link signal.
+   *
+   * This is the `a` base term in edge weighting: {{ edgeWeight(src, dst) = a + c *
+   * commonNeighbors(src, dst) }}
```
james-willis: claude says you need triple brackets for code blocks

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+  private var initialLabelCol: Option[String] = None
+  private var lgNomEntries: Int = 12
+
+  import NeighborhoodAwareCDLP.*
+
+  /**
+   * Sets weight for the neighborhood-overlap signal (common neighbors).
+   *
+   * This is the `c` term in edge weighting: {{ edgeWeight(src, dst) = a + c *
+   * commonNeighbors(src, dst) }} where `commonNeighbors(src, dst)` is the (approximate) number of
```
james-willis: claude says you need triple brackets for code blocks

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
```diff
@@ -0,0 +1,280 @@
...
+}
+
+object NeighborhoodAwareCDLP extends Logging {
+
+  val LABEL_COL = "label"
+  private val INITIAL_LABEL_COL = "initial_label"
+  private val EDGE_WEIGHT_COL = "edge_weight"
+
+  private def aggregateMessages(msgCol: Column, idType: DataType): Column = reduce(
+    collect_list(msgCol),
```
james-willis, line 214: collect list can cause OOMs when theres a lot of neighbors but the alternative is a lot of work: Catalyst native UDAF to do the reduction.

up to you if you want to do the work.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
```diff
@@ -0,0 +1,280 @@
...
+    transform(
+      map_entries(column),
+      x => struct(x.getField("value"), x.getField("key").alias("key"))))
+    .getField("key")
+
+  private def computeEdgeApproxCommonNeighbors(
+      edges: DataFrame,
+      lgNomEntries: Int,
+      neighborsScale: Double,
+      linkScale: Double): DataFrame = {
```
james-willis: in a directed graph, this defines a neighbor as one that exists on an outbound edge. 

Is this the correct definition? is there any reason to use some other definition of common neighbor?

the paper is all about undirected graphs right? so should we clarify what we do for directed graphs?


Sorry if these answer are obvious to someone with more graph experience.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: c == 0 is a valid case, it is just a regular CDLP
a == 0 is a valid case as well, it is kind of structure-based community detection

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: I will be honest: I do not know. In the paper this is just 1 without an option to change. But after some thinking I decided it would be nice to keep it flexible. We can allow > 1 but I'm not sure how will it work.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: In the classical CDLP we choose a new community as a most common across neighbors. It means that the classical CDLP threat each connection equally.

This algorithm choose a new community as a most "weighted" across neighbors where weight is a sum of edge weights to nodes from this community and the weight itself is $w_{i,j} = a + c \cdot | N_j \cap N_i |$

In the human language:
- `a` -- it is a constant that shows how important is direct connection: 0 means we ignore direct connections at all and 1.0 means we are threat connections like in the CDLP
- `b` -- it is a constant that shows how important are common neighbors: 0 means we ignore common nbrs at all and 1.0 means we consider each common neighbor equally to direct connection

There is a test case for different `a` and `c`

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: It is like a safe default.

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
SemyonSinchenko: You dislike custom UDAF as I remember because they are not working with "accelerators" (Photon, Gluten, Comet). I will be happy to use native UDAF here as well in the CLDP too. But we need to re-consider the approach to native Catalyst first.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: It is an interesting question actually. Undirected case handled in L175-182 and there is no problem (doesn't matter which one to take, src or dst). But your question about directed case is a good catch! Let me think what should we do better here: group by src and agg dst or group by dst and agg src...

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
SemyonSinchenko: There is no built-in method to get the version from spark as semver instead of string. We will end by parsing string anyway. For me it is an overcomplication.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
james-willis: But they shouldn’t BOTH be zero

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
james-willis: ok no worries

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
james-willis: Maybe we should have different parameters:

* the ratio a to c
* bool for ignore c. 
* bool for ignore a. (I'm not convinced this is a real case we need).
 
 I feel taking the weight as a ratio better models what people want - the relative weight of the direct relation and a neighbor of neighbor relation. 

Opinions?

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
james-willis: yeah sometimes.

collect_list really sucks when you really want a fold/reduce an aggregation and the cardinality of the group can be high.

really spark should have a good fold or reduce UDAF. then comet etc could just reimplement the method.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: The ration means we assume something like this $w_{i,j} = \lambda + (1 - \lambda) \cdot | N_i \cap N_j |$; I was thinking about it and it is a different story. For example, in the paper guys are trying different `c` with a constant `a = 1`

james-willis: with the $\lambda = c/a$, the equation would be  $w_{i,j} = a + \lambda \cdot | N_i \cap N_j |$ where $a \in \set{0,1}$


Aside from $a$ = 0 theres no reason to allow modifying values of $a$, because you can acomplish the exact same result with some value of $c$. so a bool of $ignoreA$ accomplishes the same thing as allowing $a$ to be set as $(0,1)$.

you actually don't need $ignoreC$ at all because it is equivalent to $\lambda =0$

SemyonSinchenko: So what would be your suggestion? I do not like the ratio idea tbh. We can remove the `a` at all and have a formula from the paper with only `c`. But I would keep the current flow. I like the "sklearn-way": everything is configurable but you won't touch most of parameters because they have safe defaults. I like the idea of just having both `a` and `c` with the default `a = 1`

james-willis: my proposal is to expose `setC: float` and `ignoreA: bool` setC solves for all relative weights of a and c. ignoreA makes a = 0

but I will not block this PR if you disagree and want to keep setA and setC.

SemyonSinchenko: And in this case we can have ìgnoreDirectLinksˋ instead of ìgnoreAˋand something like ˋstructuralSimilarityMultiplierˋ instead of ˋcˋ. What do you think?

james-willis: `ignoreDirectLinks` is a good name. 

ˋstructuralSimilarityMultiplierˋ i am not as sure of but I do not have a better name.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: Relaxed

core/src/main/scala/org/graphframes/lib/StructureAwareLabelPropagation.scala
SemyonSinchenko: I will keep as is for now. UDAF can be generic for both CDLP and this one so I would like to postpone.

core/src/main/scala/org/graphframes/lib/NeighborhoodAwareCDLP.scala
SemyonSinchenko: Good question. I updated the directed case to make this explicit.

For directed graphs, the structural-similarity term now uses a weak neighborhood: $N_v = In_v \cup Out_v$ and $W_{i,j} = N_i \cap N_j$

So propagation still follows edge direction, but the common-neighbor signal itself is direction-insensitive. I think this is a safer default for library code than using only outbound neighbors or a stricter directed definition like $Out_n \cap In_n$, because the latter can become too sparse and unstable, especially for low-degree vertices and approximate theta-sketch intersections.

Users can still tune the effect of this term with `structuralSimilarityMultiplier`, or disable the direct-link contribution with `ignoreDirectLinks`.

docs/src/04-user-guide/06-graph-clustering.md
```diff
@@ -62,6 +62,124 @@ For `graphframes` only. By default, GraphFrames uses persistent checkpoints. The
...
+### Relationship to classical CDLP
+
+You can make this algorithm behave close to classical CDLP/LPA by using:
+
+- `ignoreDirectLinks = false`
+- `structuralSimilarityMultiplier = 0`
+
+This effectively removes structural-overlap amplification and leaves direct-link voting only.
+
+However, this is **not recommended** as a replacement for classical CDLP in production. If you need classical behavior, use the dedicated CDLP API (`labelPropagation`) because it is more optimized.
```
james-willis, line 116: should we make the new API use CDLP implementation when - `structuralSimilarityMultiplier == 0`?

james-willis:
Claude raised the following naming inconsistencies:


  Naming mismatches worth raising:

  - Entrypoint name differs across languages. Scala: GraphFrame.structureAwareLabelPropagation (singular "structure", no "neighborhood"). Python: g.neighborhood_aware_cdlp(...). The Scala class is NeighborhoodAwareCDLP. Three different names for the same algorithm — pick one.
  - Python casing is inconsistent with siblings. Existing Python public methods are camelCase (labelPropagation, pageRank, shortestPaths), but the new one is neighborhood_aware_cdlp. Should probably be neighborhoodAwareCDLP (or whatever name aligns with the Scala entrypoint after the rename above).

docs/src/04-user-guide/06-graph-clustering.md
SemyonSinchenko: I do not like it. It is over-complication and I want to be able to do changes without breaking the behavior of the 10-yo code...

SemyonSinchenko:
> * Python casing is inconsistent with siblings. Existing Python public methods are camelCase (labelPropagation, pageRank, shortestPaths), but the new one is neighborhood_aware_cdlp. Should probably be neighborhoodAwareCDLP (or whatever name aligns with the Scala entrypoint after the rename above).

We should update AGENTS.md. Overall goal is to move in the direction when the Python API is "pythonic" (multi-arg methods instead of builders, snake_case naming, etc.). We cannot just change all the existing code, but we can at least avoid adding new tech debt. See https://github.com/graphframes/graphframes/issues/713 and feel free to comment.

james-willis (approved):

PR #828, opened 2026-04-29
feat: all paths between two nodes

### LINKED ISSUE #200: How to find all paths between two specific graph nodes?
I'm looking for algorithm to find ALL paths (limited by max path length)  between vertex A and vertex B. Is there any method in GraphFrame which could be used for this?

SemyonSinchenko:
A simple wrapper over `AggregateNeighbors` for enumerating all paths between two nodes.

Close #200 

P.S. PySpark + Docs after pre-approve of the API


core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+      .setRequiredVertexAttributes(Seq(GraphFrame.ID))
+
+    edgeFilterExpression.foreach { ef =>
+      agg.setEdgeFilter(SparkShims.applyExprToCol(graph.spark, ef, "edge_attributes"))
+      agg.setRequiredEdgeAttributes(graph.edges.columns.toSeq)
+    }
+
+    agg
+      .run()
+      .select(col("path"), size(col("path")).cast("long").alias("len"))
```
james-willis: len should be `size(col("path") - 1` since its the length of the edges, not nodes.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+      .setStartingVertices(fromExpression)
+      .setMaxHops(maxPathLength)
+      .setTargetCondition(SparkShims.applyExprToCol(graph.spark, toExpression, "dst_attributes"))
+      .setStoppingCondition(
+        array_contains(col("path"), AggregateNeighbors.dstAttr(GraphFrame.ID)))
+      .addAccumulator(
+        "path",
+        array(col(GraphFrame.ID)),
+        concat(col("path"), array(AggregateNeighbors.dstAttr(GraphFrame.ID))))
+      .setRequiredVertexAttributes(Seq(GraphFrame.ID))
```
james-willis: I think since we only require id here, then toExpr will only be able to reference id.

Is that intended?

core/src/main/scala/org/graphframes/lib/AllPaths.scala
james-willis: this will fix issues with off-by-one hop counts as well.

james-willis:
Should the 0 hop path be returned? BFS would return that 0 hop path

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+    val traversalGraph = if (isDirected) {
+      graph
+    } else {
+      val edgeColumns = graph.edges.columns.toSeq
+      val reversed = graph.edges.select(
+        (Seq(
+          col(GraphFrame.DST).alias(GraphFrame.SRC),
+          col(GraphFrame.SRC).alias(GraphFrame.DST)) ++
+          edgeColumns.filterNot(c => c == GraphFrame.SRC || c == GraphFrame.DST).map(col)): _*)
+      GraphFrame(graph.vertices, graph.edges.unionByName(reversed))
```
james-willis, line 182: Should we call distinct in case the user already has a large number of bidirectional edges?
```suggestion
      val reversed = graph.edges.select(
        (Seq(
          col(GraphFrame.DST).alias(GraphFrame.SRC),
          col(GraphFrame.SRC).alias(GraphFrame.DST)) ++
          edgeColumns.filterNot(c => c == GraphFrame.SRC || c == GraphFrame.DST).map(col)): _*)
      GraphFrame(graph.vertices, graph.edges.unionByName(reversed).distinct()) // distinct in case graph already has a large number of bidirectional edges
```

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    edgeFilterExpression.foreach { ef =>
+      agg.setEdgeFilter(SparkShims.applyExprToCol(graph.spark, ef, "edge_attributes"))
+      agg.setRequiredEdgeAttributes(graph.edges.columns.toSeq)
+    }
+
+    agg
+      .run()
+      .select(col("path"), size(col("path")).cast("long").alias("len"))
+      .distinct()
```
james-willis, line 206: path is a set of vertices but ignores the edges used to get there. is two difference edges between the same 2 nodes a distinct path?

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def edgeFilter(value: Column): this.type = {
+    edgeFilterExpression = Some(value)
+    this
+  }
+
+  def edgeFilter(value: String): this.type = edgeFilter(expr(value))
+
+  def run(): DataFrame = {
+    require(fromExpression != null, "fromExpr is required.")
```
james-willis, line 167: should we add requires to check for collisions on 'len' and 'path'?

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+        array_contains(col("path"), AggregateNeighbors.dstAttr(GraphFrame.ID)))
+      .addAccumulator(
+        "path",
+        array(col(GraphFrame.ID)),
+        concat(col("path"), array(AggregateNeighbors.dstAttr(GraphFrame.ID))))
+      .setRequiredVertexAttributes(Seq(GraphFrame.ID))
+
+    edgeFilterExpression.foreach { ef =>
+      agg.setEdgeFilter(SparkShims.applyExprToCol(graph.spark, ef, "edge_attributes"))
+      agg.setRequiredEdgeAttributes(graph.edges.columns.toSeq)
```
james-willis: Is this not the default?

core/src/main/scala/org/graphframes/lib/AllPaths.scala
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.functions.concat
+import org.apache.spark.sql.functions.expr
+import org.apache.spark.sql.functions.size
+import org.apache.spark.sql.graphframes.SparkShims
+import org.graphframes.GraphFrame
+import org.graphframes.WithDirection
+
+/**
+ * Computes all simple paths between source and destination vertices.
```
james-willis, line 35: The class scaladoc is the most likely entry point for users of this API — could we expand it a bit to surface a few things that are currently only discoverable by reading the implementation?

- What "simple" means here (no repeated vertices), since that's a load-bearing semantic for what's returned vs. excluded.
- That direction is configurable via `setIsDirected(false)` for undirected traversal.
- The default `maxPathLength = 10`.

A short example block (similar to the one on `BFS`) would also help users get started without grepping through the suite.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Distinct over duplicated edges is quite an expensive operation. I would like to mark it in documentation somehow instead of unconditionally call `distinct`.

SemyonSinchenko:
> Should the 0 hop path be returned? BFS would return that 0 hop path

What is zer-hop path between two nodes? Is it a self-loop in the case both source and target are the same? Does it make any sense?

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: You are right, we should keep all the attributes here. Maybe in the future we will apply expression-parsing you made to implicitly determine required state. For now let's pray that Catalyst will be able to optimize it.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Yes it is a distinct path based on the definitions from books...

james-willis: probably need to include edges in the return path value. and also have different paths for different edges.

james-willis:
yes the 0 hop path is the self loop. maybe it is a 1-hop case...

core/src/main/scala/org/graphframes/lib/AllPaths.scala
james-willis: fair enough

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: But the edge in graphframes representation is just a `src - dst`. What will be the difference between:

1. path of vertices : `A, B, C`
2. path of edges : `A-B,B-C`

For me it the same but with the first one is much easier to work in our API

SemyonSinchenko:
> yes the 0 hop path is the self loop. maybe it is a 1-hop case...

And to get it user must specify `srcExpression === dstExpression` that would be strange. I do not see any reasons to support it tbh. For cycles detection we have a standalone algorithm. I do not see this API as an API for detecting cycles.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: We do not have unique ID on edges by the end (until user provides it as a property)

james-willis: sometimes edges have more than src and dst as fields. idk what we should do.

james-willis:
I see so since we dont support cycles we dont support loops.

megawatimjfh188-code (approved):

megawatimjfh188-code (approved):

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Documented.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: There will be a separate part of documentation. I extended the scaladoc as well.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Fixed.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Let's keep it as is for now. I do not see an easy way to handle this case as well user always can do a back join on the resulted `DataFrame`.

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: added

core/src/main/scala/org/graphframes/lib/AllPaths.scala
SemyonSinchenko: Updated.

megawatimjfh188-code (approved):

james-willis (approved):

james-willis (approved):

PR #841, opened 2026-06-02
feat: HyperANF implementation

### LINKED ISSUE #749: doc: document runtime dependencies of GraphFrames

Some vendor's environments do not provide a way to resolve runtime dependencies automatically. See #748 

In Quick Start / Installation add sections about runtime dependencies of GraphFrames.

**Component**

- [ ] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**

### LINKED ISSUE #840: feat: hyperANF
https://arxiv.org/abs/1011.5599

Built-ins from Apache DataSketches (HLL, like in the paper)

**Component**

- [ ] Scala Core Internal
- [ ] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [ ] PySpark Classic
- [ ] PySpark Connect

**Additional context**

SemyonSinchenko:
- HyperANF
- bump versions in CI matrix

Close #840 
Close #749


SemyonSinchenko:
@james-willis I will add python bindings (connect/classic) and docs after pre-approve of the API design.

SemyonSinchenko:
The final goals are HyperBALL, approximate closeness centrality, etc. All of these are just simple transformations on top of the HyoerANF. Should I add these in the same PRs or in follow-up PRs?

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+      graph.edges
+        .filter(edgesFilterExpression)
+        .select(GraphFrame.SRC, GraphFrame.DST)
+        .persist(intermediateStorageLevel)
+    var hop = 1
+
+    val hop0func = udf(HyperANF.hll(lgNomEntries))
+    var state = edges
+      .groupBy(col(GraphFrame.SRC).alias(GraphFrame.ID))
+      .agg(hll_sketch_agg(GraphFrame.DST, lgNomEntries).alias("hop_1"))
```
james-willis, line 153: Is it important to make sure we always pass the same type into the hll sketch functions? on 157 we convert to string so maybe we should do that here as well

james-willis: related to this, how does this function deal with cycles? do we need a test for this case where there is a cycle to the hop 0 node?

core/src/main/scala/org/graphframes/mixins.scala
```diff
@@ -211,3 +211,20 @@ private[graphframes] trait WithDirection {
...
+ * Helper for the sketch-based algorithms.
+ */
+private[graphframes] trait WithLgNomEntries {
+  protected var lgNomEntries: Int = 12
+
+  /**
+   * Sets the log2 of the nominal entries for the data sketch. Default is 12 (4096 entries).
+   */
+  def setLgNomEntries(value: Int): this.type = {
+    require((value >= 4) && (value <= 24), "lg_nom must be between 4 and 24, defaults to 12")
```
james-willis: HLL sketches allow 4 to 21: https://apache.github.io/datasketches-java/6.1.0/org/apache/datasketches/hll/HllSketch.html

james-willis: Theta allows 26 btw: https://apache.github.io/datasketches-java/6.1.0/org/apache/datasketches/theta/UpdateSketchBuilder.html#setLocalNominalEntries(int)

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+
+      val n_state = edges
+        .join(
+          state.select(GraphFrame.ID, s"hop_${hop - 1}"),
+          col(GraphFrame.DST) === col(GraphFrame.ID),
+          "left")
+        .groupBy(col(GraphFrame.SRC).alias(GraphFrame.ID))
+        .agg(hll_union_agg(s"hop_${hop - 1}").alias(s"hop_${hop}"))
+
+      state = state.join(n_state, GraphFrame.ID)
```
james-willis: should we persist ever?

build.sbt
james-willis, line 74: Please add org.apache.datasketches.hll.HllSketch here

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+    val hop0func = udf(HyperANF.hll(lgNomEntries))
+    var state = edges
+      .groupBy(col(GraphFrame.SRC).alias(GraphFrame.ID))
+      .agg(hll_sketch_agg(GraphFrame.DST, lgNomEntries).alias("hop_1"))
+      .select(col(GraphFrame.ID), hop0func(col(GraphFrame.ID)).alias("hop_0"), col("hop_1"))
+
+    while (hop < nHops) {
+      hop += 1
+
+      val n_state = edges
```
james-willis:
```suggestion
      val nState = edges
```

core/src/main/scala/org/graphframes/lib/HyperANF.scala
james-willis:
```suggestion
      state = state.join(nState, GraphFrame.ID)
```

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+   */
+  def run(): DataFrame = {
+    val edges =
+      graph.edges
+        .filter(edgesFilterExpression)
+        .select(GraphFrame.SRC, GraphFrame.DST)
+        .persist(intermediateStorageLevel)
+    var hop = 1
+
+    val hop0func = udf(HyperANF.hll(lgNomEntries))
```
james-willis:
```suggestion
    val hop0Func = udf(HyperANF.hll(lgNomEntries))
```

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+    var state = edges
+      .groupBy(col(GraphFrame.SRC).alias(GraphFrame.ID))
+      .agg(hll_sketch_agg(GraphFrame.DST, lgNomEntries).alias("hop_1"))
+      .select(col(GraphFrame.ID), hop0func(col(GraphFrame.ID)).alias("hop_0"), col("hop_1"))
```
james-willis:
```suggestion
      .select(col(GraphFrame.ID), hop0Func(col(GraphFrame.ID)).alias("hop_0"), col("hop_1"))
```

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+    while (hop < nHops) {
+      hop += 1
+
+      val n_state = edges
+        .join(
+          state.select(GraphFrame.ID, s"hop_${hop - 1}"),
+          col(GraphFrame.DST) === col(GraphFrame.ID),
+          "left")
+        .groupBy(col(GraphFrame.SRC).alias(GraphFrame.ID))
+        .agg(hll_union_agg(s"hop_${hop - 1}").alias(s"hop_${hop}"))
```
james-willis, line 172: hll_union_agg(s"hop_${hop - 1}") will return null if hop_n is null. we should probably handle this with a coalesce to some null sketch?

core/src/main/scala/org/graphframes/lib/HyperANF.scala
```diff
@@ -0,0 +1,160 @@
...
+   * The result will contain `hop_0`, `hop_1`, `hop_2`, ..., `hop_N`, where `N` is the configured
+   * number of hops.
+   *
+   * @param value
+   *   positive number of hops to compute
+   * @return
+   *   this HyperANF instance
+   */
+  def setNHops(value: Int): this.type = {
+    require(value > 0, "n-hops cannot be nagative or zero")
```
james-willis:
```suggestion
    require(value > 0, "n-hops cannot be negative or zero")
```

james-willis (changes requested):

build.sbt
SemyonSinchenko: Why? I mean it is a part of the Spark Runtime.

core/src/main/scala/org/graphframes/lib/HyperANF.scala
SemyonSinchenko: Tbh I don't see how it can be null. It can be empty and this is handled correctly. But how can it be null (except some vertex-id is null?) P.S. null vertex IDs are considered as an invalid graph: at the moment most of GF algorithms will just fail on null-ids and handling it is very expensive (full-scan).

SemyonSinchenko: Let me check this.

SemyonSinchenko: You are right: there can be nulls. From the other side.

This code:

```python
(
    spark
    .createDataFrame([(1, None), (1, None), (1, None)], schema="k: int, v: binary")
    .toDF("k", "v")
    .groupBy("k")
    .agg(F.hll_union_agg("v").alias("v"))
    .select(F.hll_sketch_estimate("v").alias("v"))
    .show()
)
```

returns

```
+---+
|  v|
+---+
|  0|
+---+
```

so it is not a problem actually.

SemyonSinchenko: I added a test for that case.

core/src/main/scala/org/graphframes/lib/HyperANF.scala
SemyonSinchenko: Cycles are not a problem. We are limited by hHops. When user do `union` + `estimate` all the cycles will be gone.

core/src/main/scala/org/graphframes/lib/HyperANF.scala
SemyonSinchenko: Added standard persist-checkpoint flow.

core/src/main/scala/org/graphframes/mixins.scala
SemyonSinchenko: Fixed

core/src/main/scala/org/graphframes/lib/HyperANF.scala
james-willis: well that is my concern we cannot correctly identify a cycle because the union wont dedupe (pseudo code) `hash(str(numericTypeId))` and `hash(numericTypeId))`

build.sbt
james-willis: because we depend directly on it. It is not critical but IMO it is the best practice to enumerate direct dependencies explicitly. It can encode version requirements and such

core/src/main/scala/org/graphframes/lib/HyperANF.scala
SemyonSinchenko: @james-willis I fixed the typing problem: now all the hll have the same type

james-willis (approved):

james-willis (approved):

PR #846, opened 2026-06-08
Optimization: Add pruning for connected component 2 phase algorithm

WeichenXu123:

Optimization: Add pruning for connected component 2 phase algorithm.

boost performance

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/TwoPhase.scala
+++ b/core/src/main/scala/org/graphframes/lib/TwoPhase.scala
@@ -566,7 +566,7 @@ private[graphframes] object TwoPhase extends Logging {
       // For the dense graph, its edges will be pruned at each large/small star join iteration,
       // and we will try the optimization once the graph becomes sparse.
       if ((edgeCnt < sparsityThreshold * numNodes) && (edgeCnt > 0)
-         && (iteration >= optStartIter) && (!triedToOptimize)) {
+        && (iteration >= optStartIter) && (!triedToOptimize)) {
         edgesBeforePruning = ee
         pruneLeafNodes(ee, intermediateStorageLevel, numNodes, shrinkageThreshold) match {
           case Some(r) =>
--- a/core/src/main/scala/org/graphframes/lib/TwoPhase.scala
+++ b/core/src/main/scala/org/graphframes/lib/TwoPhase.scala
@@ -110,14 +110,15 @@ private[graphframes] object TwoPhase extends Logging {
   }
 
   /**
-   * Computes the sum of all `min_nbr` values in the given DataFrame, cast to DecimalType(38, 0)
-   * for high precision. Used to detect convergence between iterations.
+   * Computes the sum of all `min_nbr` values and the row count of the given DataFrame in a single
+   * Spark job. The sum is cast to DecimalType(38, 0) for high precision. Used to detect
+   * convergence between iterations and to check graph sparsity for the pruning optimization.
    */
-  private def calcMinNbrSum(minNbrsDF: DataFrame): BigDecimal = {
-    minNbrsDF
-      .select(sum(col(MIN_NBR).cast(DecimalType(38, 0))))
+  private def calcMinNbrSum(minNbrsDF: DataFrame): (BigDecimal, Long) = {
+    val row = minNbrsDF
+      .select(sum(col(MIN_NBR).cast(DecimalType(38, 0))), count("*"))
       .first()
-      .getAs[BigDecimal](0)
+    (row.getAs[BigDecimal](0), row.getLong(1))
   }
 
   /**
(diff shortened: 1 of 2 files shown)


WeichenXu123:
This is an optimization that has been implemented in Databricks internal graphframe lib. Now we decide to contribute it to OSS. CC @SemyonSinchenko

SemyonSinchenko:
First of all, thanks a lot for the contribution @WeichenXu123 !

I run local benchmarks on one of small LDBC' graphs. The first results are very promising ❤️‍🔥❤️‍🔥❤️‍🔥

Wiki-Talk, 2M/5M -- the same is run in CI and published on website.
Spark 4.0.2

**This PR**
```
[info] Result "org.graphframes.benchmarks.ConnectedComponentsBenchmark.benchmarkConnectedComponents":
[info]   87.427 ±(99.9%) 38.537 s/op [Average]
[info]   (min, avg, max) = (85.203, 87.427, 89.406), stdev = 2.112
[info]   CI (99.9%): [48.890, 125.964] (assumes normal distribution)
```

**Main**
```
[info] Result "org.graphframes.benchmarks.ConnectedComponentsBenchmark.benchmarkConnectedComponents":
[info]   129.323 ±(99.9%) 32.521 s/op [Average]
[info]   (min, avg, max) = (127.966, 129.323, 131.341), stdev = 1.783
[info]   CI (99.9%): [96.802, 161.843] (assumes normal distribution)
```

**On the defaults in GraphFrames**

GraphFrames is very old project. And ConnectedComponents is a widely used algorithm. So, we are trying not to change defaults in the minor-version updates and we are considering changing defaults as a breaking changes. At the same time, a couple of versions ago I was experimenting with `skewedJoin` in CC and found that even on the graphs with gigantic component AQE handles skew better than manual `collect` + `broadcast`: that is the note in documentation ["AQE-mode"](https://graphframes.io/04-user-guide/05-traversals.html#aqe-broadcast-mode).

So, we have actually two different implementations of the `two_phase`: `run` and `runAQE`. Can I ask you to put the same leaf-nodes optimization to the `runAQE` as well? Because at the moment it is still 4x time faster than `run` with leaf-node optimization (and 6x faster than master)?

```
[info] Result "org.graphframes.benchmarks.ConnectedComponentsBenchmark.benchmarkConnectedComponents":
[info]   20.206 ±(99.9%) 12.791 s/op [Average]
[info]   (min, avg, max) = (19.401, 20.206, 20.685), stdev = 0.701
[info]   CI (99.9%): [7.415, 32.998] (assumes normal distribution)
```

P.S.
To reproduce:
```
sbt:graphframes> benchmarks/Jmh/run -p graphName=wiki-Talk -p algorithm=two_phase -p broadcastThreshold="10000" org.graphframes.benchmarks.ConnectedComponentsBenchmark
```

and to run in "AQE-mode":
```
sbt:graphframes> benchmarks/Jmh/run -p graphName=wiki-Talk -p algorithm=two_phase -p broadcastThreshold="-1" org.graphframes.benchmarks.ConnectedComponentsBenchmark
```

And thanks a again for this contribution! The idea is brilliant and the results look very promising!

core/src/main/scala/org/graphframes/lib/TwoPhase.scala
```diff
@@ -288,8 +372,44 @@ private[graphframes] object TwoPhase extends Logging {
...
+
+        // According to such heuristic rule, we can determine when and whether we should
+        // perform the optimization. For the sparse graphs (defined by sparsityThreshold),
+        // we will try such optimization at the end of $optStartIter iteration (default is 2).
+        // For the dense graph, its edges will be pruned at each large/small star join iteration,
+        // and we will try the optimization once the graph becomes sparse.
+        if ((edgeCnt < sparsityThreshold * numNodes) && (edgeCnt > 0)
+          && (iteration >= optStartIter) && (!triedToOptimize)) {
+          edgesBeforePruning = ee
+          pruneLeafNodes(ee, intermediateStorageLevel, numNodes, shrinkageThreshold) match {
```
SemyonSinchenko, line 397: Should we update the `currSum` here? I mean if we shrink the graph and convergence should happen next iteration we won't catch that algorithm is converged because the `preSum` at the next iteration will be `currSum` from the iteration when shrinking happened but that sum was computed before shrinking and it is always be bigger.

WeichenXu123: This is a trade-off, although convergence detection is delayed by one iteration, it saves the time for recomputing calcMinNbrSum on the shrunken ee after pruning

SemyonSinchenko: From what I see in the code, `minNbrs1` is persisted at this moment as well the `r._1` is. Both are relatively small compared to edges. We just need to do a left-anti join between two already persisted `DataFrame`s and run one `sum` aggregation. It should not impact the overall performance. Something like:

```scala
if (shrunkRound) {
  val (currSum, _) = calcMinNbrSum(
    minNbrs1.join(shrunkenGraphNodes, Seq(GraphFrame.ID), "left-anti"))
  shrunkRound = false
}
```
at the line after the convergence checking for the current round.

From the other side, the full iteration is a few joins with edges (even shrunk) and aggregations. My concern is we are changing the default behavior and some users may face performance degradation in the case they are so unlucky that faced exactly the worsen case when convergence would be exactly one iteration after shrunk...

SemyonSinchenko: I'm not 100% sure to be honest because convergence next iteration after the shrunk should be a very rare case actually. Let's keep it as is, keep it in mind and be ready to make a patch.

SemyonSinchenko (approved):
I see a clear win on local benchmarks as well I do not see any risk except a very rare case of potential risk of one additional iteration. But the last one should be a) very rare b) on top of the already shrunk graph c) should not affect overall performance

SemyonSinchenko:
Thanks for the contribution @WeichenXu123 !

james-willis:
Thanks @WeichenXu123! Hoping this is the start of getting Databricks back to having modern Graphframes version.

WeichenXu123:
quick question @SemyonSinchenko @james-willis 
when do you plan to release a new Graphframe OSS version ? then Databricks ML runtime can pre-install it (instead of current preinstalled old version )

SemyonSinchenko:
@WeichenXu123 I created a tracking issue #856

PR #847, opened 2026-06-09
feat: pregel memory optimizations

### LINKED ISSUE #835: feat: requiredEdgeColumns API in Pregel
At the moment all the edge attributes are packed to the struct and persisted (!):

```scala
    val edges = graph.edges
      .select(col(SRC).alias("edge_src"), col(DST).alias("edge_dst"), struct(col("*")).as(EDGE))
      .repartition(col("edge_src"))
      .persist(intermediateStorageLevel)
```

While in built-in algorithms we are work-arounding it by explicitly select only required columns, it would be nice to add an API for end users that allows to specify required columns.

`requiredEdgeColumns`: if specified we are selecting only required edge columns. If it is empty we should remove `struct(col("*")).as(EDGE)` from edges at all. At the moment it is more like a bug / unclear behaviour: it always add a struct with `SRC` and `DST` that is persisted (!). It is bad.

**Component**

- [x] Scala Core Internal
- [x] Scala API
- [x] Spark Connect Plugin
- [ ] Infrastructure
- [x] PySpark Classic
- [x] PySpark Connect

**Additional context**
While it may look like a breaking change, for me it is more like fixing unspecified behavior on a very-very rare case someone (wrongly) uses `Pregel.edge(SRC)` instead of `Pregel.src(ID)`. In the case `requiredEdgeColumns` is empty I would like to drop `EDGE` at all from `edges`.

By default the `requiredEdgeColumns` should use all the edge columns except the `SRC` and `DST`. AND in the case there are no additional columns, `EDGE` struct should not be created.

We may mention it in release notes for the rare case someone is still using the old `EDGE` for any reason (that is very-very unlikely imo).

SemyonSinchenko:
Edge columns are not persisted by default anymore.

x2 less peak memory usage + small benefit on local benchmarks:

**This PR**
```
[info] Result "org.graphframes.benchmarks.LabelPropagationBenchmark.benchmarkLabelPropagation":
[info]   40.889 ±(99.9%) 14.652 s/op [Average]
[info]   (min, avg, max) = (39.969, 40.889, 41.450), stdev = 0.803
[info]   CI (99.9%): [26.237, 55.540] (assumes normal distribution)
```

**Main**
```
[info] Result "org.graphframes.benchmarks.LabelPropagationBenchmark.benchmarkLabelPropagation":
[info]   41.504 ±(99.9%) 12.715 s/op [Average]
[info]   (min, avg, max) = (40.712, 41.504, 42.025), stdev = 0.697
[info]   CI (99.9%): [28.789, 54.219] (assumes normal distribution)
```

Close #835 

P.S. I do not see this as a breaking change. It was not documented anywhere that edges are persisted by default. I see it as a small change that will benefit 99% of users and only ~1% that used edge-columns in `Pregel` should do a small update.


SemyonSinchenko:
PySpark bindings and docs update after pre-approve

james-willis (approved):

james-willis (approved):

PR #865, opened 2026-06-17
feat: path compaction

### LINKED ISSUE #827: feat: non-branching paths compaction

At the moment there is no single API for graph simplification by compaction of the non-branching paths.

An API for simplification of graphs that do non-branching paths compaction.

1. Users should be able to specify the cycles policy
2. Users should be able to specify any amount of custom aggregations
3. Users should be able to specify the expression for "always keep" nodes/edges
4. Users should be able to specify the expression for "always compact" nodes/edges

**Component**

- [x] Scala Core Internal
- [x] Scala API
- [ ] Spark Connect Plugin
- [ ] Infrastructure
- [x] PySpark Classic
- [x] PySpark Connect

**Additional context**

SemyonSinchenko:
The new paths compaction API; only core -- public API will follow after pre-approve

Close #827


james-willis:
I'm not sure that the 'indegree == outdegree == 1' is quite the right rule depending on the graph.

like an undirected graph should have 2 distinct undirected edges. so `a <-> b <-> c` should be compacted to `a<->c` but wouldnt b be considered to have an in-degree of 2 here?

This is still true in a directed graph: `a <-> b <-> c -> d` I would say the correct result would be `a<->c->d`.

james-willis:
Similarly I think dead ends would be incorrectly eliminated: `...a<->b` should not eliminate b even though it has one in and one out in the directed graph.

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+   * appears in `value` it is silently removed.
+   *
+   * @param value
+   *   vertex attribute column names to carry into the aggregation input
+   * @return
+   *   this instance for method chaining
+   */
+  def requiredVertexAttributes(value: Seq[String]): this.type = {
+    requiredVertexAttributes = value.filter(_ != GraphFrame.ID).toSeq
+    this
```
james-willis, line 97: you need to provide these columns for always remove/always keep filters as well

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+   */
+  def setDropCompactedCycles(value: Boolean): this.type = {
+    dropCompactedCycles = value
+    this
+  }
+
+  /**
+   * Register a named aggregation expression to be evaluated over the interior vertices of each
+   * compacted path. The expression is applied inside a `groupBy(component).agg(...)`; it may
+   * reference any vertex attribute column of the original graph as well as `orig_src` and
```
james-willis, line 114: I think it needs to be a column saved by requiredVertexAttributes

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+    val annotated = ({
+      val inner = candidates
+        .join(
+          persistedBaseEdges.select(col(GraphFrame.SRC).as("orig_src"), col(GraphFrame.DST)),
+          col(GraphFrame.DST) === col(GraphFrame.ID),
+          "inner")
+        .join(
+          persistedBaseEdges.select(col(GraphFrame.SRC), col(GraphFrame.DST).as("orig_dst")),
+          col(GraphFrame.SRC) === col(GraphFrame.ID),
+          "inner")
```
james-willis, line 215: should we drop the src and dst from the join here? I think theyre dupes of the id

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+ *   import org.apache.spark.sql.functions.collect_set
+ *
+ *   new PathCompaction(graph)
+ *     .setDropCompactedCycles(false)
+ *     .withAggExpression("ids", collect_set("id"))
+ *     .run()
+ * }}}
+ *
+ * The returned `DataFrame` has columns `src`, `dst`, `count`, followed by one column per
+ * additionally registered aggregation expression.
```
james-willis, line 67: I dont think the Map will give us the ordering semantics youre claiming here. need some kind of order map implementation for this

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+        size(col("compacted_src")) === 0 && size(col("compacted_dst")) === 0)
+      .withColumn(
+        GraphFrame.SRC,
+        when(col("is_cycle"), array_min(col("inner_ids")))
+          .otherwise(element_at(col("compacted_src"), 1)))
+      .withColumn(
+        GraphFrame.DST,
+        when(col("is_cycle"), array_min(col("inner_ids")))
+          .otherwise(element_at(col("compacted_dst"), 1)))
+
```
james-willis, line 285: We need to do `count = count - 1 if is_cycle` because we keep one of the candidates

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+   *   this instance for method chaining
+   */
+  def requiredVertexAttributes(value: Seq[String]): this.type = {
+    requiredVertexAttributes = value.filter(_ != GraphFrame.ID).toSeq
+    this
+  }
+
+  /**
+   * If `true` (default), cycles of candidates are not compacted: they are absent from the set of
+   * compacted edges, so their original edges survive unchanged in the output. If `false`, each
```
james-willis, line 102: dont we drop all columns other than src,dst?

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
```diff
@@ -0,0 +1,345 @@
...
+
+    // clean-up
+    persistedBaseEdges.unpersist(blocking = true)
+    annotated.unpersist(blocking = true)
+    interiorEdges.unpersist(blocking = true)
+    components.unpersist(blocking = true)
+    compactedInteriorIds.unpersist(blocking = true)
+
+    resultIsPersistent()
+    allEdges
```
james-willis, line 337: should we return a graphframe instead?

james-willis (commented):
initial quick review with AI. will come back later for a more complete review

core/src/main/scala/org/graphframes/spatial/PathCompaction.scala
SemyonSinchenko: Please no 😂

There is one API (old PageRank) that returns GraphFrame and I have no idea what do with it. It is (almost) impossible to support anything that returns GraphFrame in Spark-Connect.

PR #868, opened 2026-06-24
feat: GQL-2 -- grammar, plugin, GqlAst, AstBuilder

SemyonSinchenko:


SemyonSinchenko:
Tests fails here are related to the functions (lowering) that is a part of the QueryExecution. Other tests are fine.

PR #869, opened 2026-06-24
feat: GQL-3 -- function registry

SemyonSinchenko:


SemyonSinchenko:
Compile should work after we merge #868

PR #578, opened 2025-04-15
Create Pregel and Neo4j Ingeration tutorials and add Pregel API to user guide

rjurney:

PR #877, opened 2026-07-28
Add pre-commit and Python >= 3.12 numpy dependencies to Poetry

rjurney:

I had numpy errors in Python 3.13/3.14 and they use a different version of numpy from pandas being a greater version and dropping Python <= 3.12 support. This fixes those errors with a conditional numpy dependency.

Also, pre-commit was not a dev dependency and it should be.

These changes are needed for my Pregel tutorial to work. Aside from this, it is now complete! I ran it end-to-end. See #578


python/pyproject.toml
```diff
@@ -51,6 +51,12 @@ version = "0.0.0"
 poetry-dynamic-versioning = { version = ">=1.0.0,<2.0.0", extras = ["plugin"] }
 poetry-plugin-export = ">=1.8"
 
+[tool.poetry.dependencies]
+pyspark = ">=3.5, <4.2"
```
james-willis, line 55: why dont we allow spark 4.2?

SemyonSinchenko: Why are we adding pyspark dependency while we are considered it as "provided"?

python/pyproject.toml
```diff
@@ -51,6 +51,12 @@ version = "0.0.0"
...
+numpy = [
```
SemyonSinchenko, line 56: Why are we adding numpy dependency?

SemyonSinchenko:
> I had numpy errors in Python 3.13/3.14 and they use a different version of numpy from pandas being a greater version and dropping Python <= 3.12 support. This fixes those errors with a conditional numpy dependency.

Could you clarify please? We do not use pandas anyhow in the graphframes code.

> These changes are needed for my Pregel tutorial to work.

I would prefer to achieve the same by just introducing a dependency group.

rjurney:
This was a bug I encountered while upgrading Java from 17 to 21 and Spark from 3.5 to 4.1.3. It isn't around any longer. Closing.

PR #893, opened 2026-08-28
fix: map randomized_contraction output back to non-integral vertex IDs

### LINKED ISSUE #892: randomized_contraction fails on non-integral vertex IDs (CAST_INVALID_INPUT)
## Summary

`ConnectedComponents` with `algorithm = "randomized_contraction"` fails on any graph whose vertex IDs are not integral, unless `useLabelsAsComponents` is set. The failure is a Spark cast error, not a GraphFrames-level message.

This contradicts the user guide, which states:

> Unlike `two_phase`, `randomized_contraction` **always** produces random `Long` component IDs regardless of the input vertex ID type, unless `use_labels_as_components=True` is set.

`two_phase` handles non-integral IDs correctly, so this is specific to `randomized_contraction`.

## Cause

At the end of `RandomizedContraction.run`, the `useLabelsAsComponents = false` branch maps results back with:

```scala
inputGraph.vertices
  .join(finalReps, ID, "left")
  .select(
    col(ID),
    coalesce(col(ConnectedComponents.COMPONENT), col(ID))
      .alias(ConnectedComponents.COMPONENT))
```

`COMPONENT` is a `Long` produced by the contraction; `ID` is the *original* vertex id. When the original id is a `String`, Spark coerces the two `coalesce` branches to a common type and casts the id to `BIGINT`. For any non-numeric id that cast fails.

The `ORIG_ID` branch above it does the same thing with `col(ORIG_ID)`.

Note the graph preparation itself is fine — `prepare` re-indexes through `graph.indexedVertices`, which correctly uses `monotonically_increasing_id()` for non-integral ids rather than casting. The problem is only in the final fallback, which reaches back to the un-indexed id.

## Reproduction

Verified locally against `io.graphframes:graphframes-spark4_2.13:0.12.2` on Apache Spark 4.0.1 / Scala 2.13 / Java 17.

```scala
val vertices = spark.createDataFrame(Seq("a", "b", "c").map(Tuple1.apply)).toDF("id")
val edges = spark.createDataFrame(Seq(("a", "b"))).toDF("src", "dst")
val g = GraphFrame(vertices, edges)

// works
g.connectedComponents.setAlgorithm("two_phase").setUseLabelsAsComponents(false).run().show()

// throws
g.connectedComponents.setAlgorithm("randomized_contraction").setUseLabelsAsComponents(false).run().show()
```

`two_phase` returns `Long` components as documented:

```
+---+-------------+
| id|    component|
+---+-------------+
|  a|1382979469312|
|  b|1382979469312|
|  c|1047972020224|
+---+-------------+
```

`randomized_contraction` on the same graph fails:

```
[CAST_INVALID_INPUT] The value 'b' of the type "STRING" cannot be cast to "BIGINT"
because it is malformed. Correct the value as per the syntax, or change its target
type. Use `try_cast` to tolerate malformed input and return NULL instead. SQLSTATE: 22018
```

Originally hit in the wild while evaluating `randomized_contraction` for Apache Sedona's DBSCAN, which assigns vertex ids as `sha2(to_json(struct("*")), 256)`. Every DBSCAN test failed there with the hex hash in the cast message (apache/sedona#3299), which is what led to the minimal case above.

## Test coverage gap

`randomized_contraction` appears in the test suite only in `core/src/test/scala/org/graphframes/ldbc/TestLDBCCases.scala:214`, whose graphs are explicitly `LongType`:

```scala
.schema(StructType(Seq(StructField("src", LongType), StructField("dst", LongType))))
```

`ConnectedComponentsSuite` covers `String` ids, but only for `graphx` and `graphframes`/`two_phase` — for example the "friends graph with different broadcast thresholds" case iterates `Seq(("graphx", 1000000), ("graphframes", 100000), ("graphframes", 1), ("graphframes", -1))`. Adding `randomized_contraction` to the String-id cases would catch this.

## Impact

The docs currently recommend `randomized_contraction` as the best available algorithm, so users following that advice on `String`-keyed graphs hit a cast error whose message points at Spark rather than at the algorithm choice. Worth considering a clear error if it is meant to be integral-only — but per the docs it is meant to work.

james-willis:
Fixes #892.

## Problem

`randomized_contraction` produces wrong results, or fails outright, on graphs whose vertex IDs are not integral — despite the user guide stating it "always produces random `Long` component IDs regardless of the input vertex ID type".

`finalReps` is keyed by the internal `LongType` vertex index assigned in `prepare()` (via `graph.indexedVertices`), **not** by the original vertex ID. The `useLabelsAsComponents = false` branch nevertheless joined `inputGraph.vertices` on `ID` and fell back to `coalesce(component, id)`. Both the join key and the coalesce mix `StringType` with `LongType`, so Spark's implicit coercion decides the outcome:

- **Spark 4 (ANSI on)** — the string side is cast to `BIGINT` and the query fails with `CAST_INVALID_INPUT`.
- **Spark 3.5 (ANSI off)** — no exception. The join silently never matches, every vertex falls through to the fallback, and `component` comes back as `StringType`. This is the more dangerous case: wrong answers, no error.

The `useLabelsAsComponents = true` path already handled this correctly, via a dedicated `!hasIntegralIdType` branch. The `false` path simply never got one.

## Fix

Add the missing branch, mirroring the existing correct one: map back through `indexedVertices`, join `LONG_ID` to `LONG_ID`, and use `LONG_ID` as the coalesce fallback for isolated vertices so both sides are `LongType`. Purely additive — no existing branch is modified, and the integral-ID path is untouched.

`GraphFrame.indexedVertices` is a `lazy val`, so `prepare()` and this branch observe the same `monotonically_increasing_id()` assignment; the index mapping is stable across both uses. The `isGraphPrepared = true` path is unaffected, since a prepared graph has integral IDs and still takes the final `else`.

## Tests

`randomized_contraction` was previously exercised only by `TestLDBCCases`, whose graphs are explicitly `LongType` — which is why this shipped. Added `ConnectedComponentsSuite` coverage over `Graphs.friends` (String IDs, with `g` isolated so the coalesce fallback is exercised) for both `useLabelsAsComponents` values, asserting component grouping and column type.

Verified locally: `testOnly *ConnectedComponentsSuite` 65 passed; full `core/test` 332 passed with 14 pre-existing Spark-version cancellations; `scalafmtCheckAll` and `scalafixAll --check` clean. Before the fix, the new `useLabelsAsComponents=false` case fails on the default 3.5.8 build with `Column component must be of type LongType but was actually StringType`.

End-to-end on Spark 4.0.1: pre-fix `CAST_INVALID_INPUT`; post-fix `struct<id:string,component:bigint>` with correct grouping and a `Long` component for the isolated vertex.

I did not run the full multi-Spark matrix from AGENTS.md locally — relying on CI for that.

## Two related issues found but deliberately not fixed here

1. **The zero-edge early return has the same class of bug.** It returns `inputGraph.vertices.select(col(ID), col(ID).alias(COMPONENT))` unconditionally, so a String-ID graph with no edges yields `component: string` (and an Int-ID graph yields `component: int`), where `two_phase` yields `bigint`. After this PR, `randomized_contraction` on a String graph returns `bigint` when edges exist and `string` when none do. Left alone because fixing it changes behaviour for integral-ID graphs too; happy to fold it in or open a follow-up.

2. **`randomized_contraction` drops all vertex attributes.** Every output branch selects only `id` and `component`, whereas `two_phase` and `graphx` also return `col("attr.*")`. This contradicts the `ConnectedComponents` scaladoc ("The resulting DataFrame contains all the vertex information and one additional column") and affects all ID types, so it is out of scope here. Same root cause for going unnoticed: the LDBC tests compare only `id` and `component`.

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/RandomizedContraction.scala
+++ b/core/src/main/scala/org/graphframes/lib/RandomizedContraction.scala
@@ -253,6 +253,20 @@ private[graphframes] object RandomizedContraction extends Logging with Serializa
             col(ID),
             coalesce(col("new_component"), col(ID))
               .alias(ConnectedComponents.COMPONENT))
+      } else if (!inputGraph.hasIntegralIdType) {
+        // `finalReps` is keyed by the internal LongType vertex index, not by the original
+        // vertex ID, so for non-integral ID types the results have to be mapped back through
+        // `indexedVertices`. Joining `inputGraph.vertices` on ID directly would make Spark
+        // coerce the two sides to a common type, which either fails with CAST_INVALID_INPUT
+        // (ANSI mode) or silently produces wrong results.
+        // See https://github.com/graphframes/graphframes/issues/892
+        inputGraph.indexedVertices
+          .withColumnRenamed(ID, ConnectedComponents.ORIG_ID)
+          .join(finalReps, col(ID) === col(LONG_ID), "left")
+          .select(
+            col(ConnectedComponents.ORIG_ID).alias(ID),
+            coalesce(col(ConnectedComponents.COMPONENT), col(LONG_ID))
+              .alias(ConnectedComponents.COMPONENT))
       } else {
         inputGraph.vertices
           .join(finalReps, ID, "left")
--- a/core/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
+++ b/core/src/test/scala/org/graphframes/lib/ConnectedComponentsSuite.scala
@@ -97,6 +97,40 @@ class ConnectedComponentsSuite extends SparkFunSuite with GraphFrameTestSparkCon
     }
   }
 
+  Seq(true, false).foreach { useLabelsAsComponents =>
+    test(
+      "randomized contraction with non-integral IDs, " +
+        s"useLabelsAsComponents=$useLabelsAsComponents") {
+      // See https://github.com/graphframes/graphframes/issues/892 : the randomized contraction
+      // implementation works on the internal LongType vertex indices, so the results have to be
+      // mapped back onto the original (here: StringType) vertex IDs.
+      val friends = Graphs.friends
+      val components = friends.connectedComponents
+        .setAlgorithm(ConnectedComponents.ALGO_RANDOMIZED_CONTRACTION)
+        .setUseLabelsAsComponents(useLabelsAsComponents)
+        .run()
+      TestUtils.checkColumnType(components.schema, ID, DataTypes.StringType)
+      val expected = Set(Set("a", "b", "c", "d", "e", "f"), Set("g"))
+
+      if (useLabelsAsComponents) {
+        TestUtils.checkColumnType(components.schema, "component", DataTypes.StringType)
+        assertComponents(components, expected)
+      } else {
+        // Component IDs are arbitrary Longs in this mode, so only the grouping is checked.
+        TestUtils.checkColumnType(components.schema, "component", DataTypes.LongType)
+        val grouping = components
+          .select(col("component"), col(ID))
+          .collect()
+          .groupBy(_.getLong(0))
+          .values
+          .map(_.map(_.getString(1)).toSet)
+          .toSet
+        assert(grouping === expected)
+      }
+      components.unpersist()
+    }
+  }
+
   Seq(true, false).foreach(useSkewedJoin => {
     Seq(true, false).foreach(useLocalCheckpoint => {
       val testPostfixName = s"${if (useLocalCheckpoint) " with local checkpoint"


SemyonSinchenko (approved):

core/src/main/scala/org/graphframes/lib/RandomizedContraction.scala
```diff
@@ -253,6 +253,20 @@ private[graphframes] object RandomizedContraction extends Logging with Serializa
             col(ID),
             coalesce(col("new_component"), col(ID))
               .alias(ConnectedComponents.COMPONENT))
+      } else if (!inputGraph.hasIntegralIdType) {
```
james-willis, line 256: unclear if the join or maintaining the column through calculation would be faster.

Issue #894, opened 2026-08-29
RandomizedContraction can leave checkpoint directories behind after a failed run


SauronShepherd:
### Problem

RandomizedContraction creates a per-run checkpoint directory and deletes it only after the output has been materialized successfully. Its finally block currently unpersists DataFrames and deregisters _axpb, but does not remove the temporary checkpoint directory.

If an intermediate operation fails, checkpoint data can remain in the filesystem or a remote object store. Repeated executions may therefore leak temporary storage.

This is related to the historical cleanup concern in #162, but specifically covers failed executions in the current RandomizedContraction implementation.

### Proposed solution

- Reuse a scoped cleanup routine for the per-run checkpoint directory.
- Run it on both successful completion and from finally.
- Do not mask the original failure if cleanup itself cannot complete; log a warning instead.
- Add a regression test that fails after intermediate parquet materialization and verifies that no randomized-contraction-* directory remains.

The change does not modify public APIs.

SemyonSinchenko:
Fixed by #895

PR #895, opened 2026-08-29
fix: clean randomized contraction checkpoints on failure

SauronShepherd:

Ensure that RandomizedContraction checkpoint directories are cleaned up on both successful and failed executions.

A scoped cleanup routine is now called from the normal completion path and from finally. Cleanup failures are logged without masking the original algorithm failure.

A regression test forces a failure after intermediate parquet materialization and verifies that no randomized-contraction-* directory remains.

If an intermediate Spark or filesystem operation fails, the current implementation can leave temporary checkpoint data behind. Repeated executions can therefore leak storage, especially when using remote object stores.

This addresses #894.

### Validation

- RandomizedContractionSuite on WSL with Java 11
- 14 tests passed
- 0 failed
- 0 aborted
- git diff --check passed

The change does not modify public APIs.

### DIFF
--- a/core/src/main/scala/org/graphframes/lib/RandomizedContraction.scala
+++ b/core/src/main/scala/org/graphframes/lib/RandomizedContraction.scala
@@ -21,6 +21,7 @@ import java.io.IOException
 import java.util.UUID
 import scala.collection.mutable.Stack
 import scala.util.Random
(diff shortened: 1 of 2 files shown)


SemyonSinchenko (approved):
LGTM, thanks for the contribution!

PR #896, opened 2026-08-31
ci: run python lint on tests/ and docs/, as its own job

james-willis:
## Why

Formatting churn keeps showing up inside unrelated feature PRs (most recently in #809, in a 5k+ line docs PR). The usual conclusion is "CI doesn't lint Python" — but it does: `python-ci.yml` has run `black --check`, `flake8` and `isort --check` for a while, and it is green on main.

The problem is that the check has holes.

| Path | .py files | Linted before | Linted after |
|---|---|---|---|
| `python/graphframes` | 23 | yes | yes |
| `python/dev`, `dev` | 5 | yes | yes |
| **`python/tests`** | **5** | **no** | **yes** |
| **`python/docs`** | **3** | **no** | **yes** |

`python/tests` is the one that hurts: it is touched by most feature PRs, and it has never been format-checked. So drift lands on main quietly and then surfaces as churn in whoever's PR happens to touch those files next.

Reproduced against `main` with the repo's own pinned tools:

```
# old CI paths
black --check --config pyproject.toml graphframes dev ../dev
  -> All done! 31 files would be left unchanged.        # passes

# same tree, new CI paths
black --check --config pyproject.toml graphframes tests docs dev ../dev
  -> 6 files would be reformatted.                      # catches it
```

## What this changes

- **Cover `tests/` and `docs/`** in both `python-ci.yml` and `.pre-commit-config.yaml`, with a comment on each noting the two path lists must stay in sync. The pre-commit config had the identical gap, so local hooks did not catch it either.
- **Move the style check into its own `lint` job.** It was a step inside the Spark test matrix, so it ran three times (once per matrix entry) and only after a full JDK + Spark + jar-build setup. It now runs once, independently, with no Spark, and reports its own status. black/isort/flake8 are separate steps so a red check names the tool.
- **Fix `[tool.black] include`**, which was `["graphframes"]` — a list where black expects a regex string. A list silently disables black's default `\.pyi?$` filter, so black tried to parse `Makefile`, `.rst`, `.css` and `.js` when handed a directory. This is why `docs/` could not simply be added to the path list.
- **Exclude the generated protobuf stubs** (`graphframes/connect/proto/`) from black and isort. They are emitted by `buf generate` (see `buf.gen.yaml`), so formatting them fights the generator — black reflows the output, the next regen emits the original, and the churn returns. #809 shows this: `graphframes_pb2.pyi` picked up a reformatted `__slots__` tuple unrelated to that PR. black needs `force-exclude` rather than `exclude`, since `exclude` is not applied to paths passed explicitly on the command line, and both CI and the pre-commit hook pass paths explicitly. flake8 is left as-is — `tox.ini` already carves out `E501`/`F401` there and its remaining checks do not rewrite the file.
- **Apply black + isort** to the newly covered files and fix the 6 remaining flake8 findings (3x `E265`, 2x `E501`, 1x `W391`).

Nothing outside `python/tests` and `python/docs` is reformatted. `graphframes`, `dev` and `../dev` were already clean and are byte-identical.

## On black vs ruff

I looked at swapping to ruff while here and deliberately did not. The defect is coverage, not tooling — ruff would not have caught any of this either, since it would have been pointed at the same paths. For the record, `ruff format` over the code black already keeps clean is only ~36 lines across 6 files, so the migration is cheap and worth discussing, just not bundled into the PR that is supposed to *stop* incidental reformatting.

## One thing this does not fix

`black --check` structurally cannot catch the churn in `graphframes_client.py` in #809. That one is black's **magic trailing comma**:

- the original signature line is **90 chars** — between black's default `88` and this repo's configured `100`
- something formatted it at 88 (an editor format-on-save typically formats a temp buffer, where black finds no `pyproject.toml` and falls back to its default). Black split it one-arg-per-line **and added a trailing comma**
- black treats that comma as an explicit "keep this exploded" marker, so a correct run at 100 **preserves** it rather than collapsing it back

Both forms therefore pass `black --check` at line-length 100, and CI is green on either. I confirmed it is not version drift: black 26.5.1 at 100 also leaves the original untouched. It is a one-way ratchet — one run under the wrong config leaves a mark the right config never undoes.

`black --check --skip-magic-trailing-comma` does catch it, but adopting that repo-wide is a 266-line reformat across 17 files and would flatten intentionally-exploded collections. That is a separate, opinionated call — happy to open it as a follow-up if there's appetite. Making the repo's line-length discoverable from the root (or documenting `--config`) would also close the hole at the source.

This is intended as the scoped, standalone reformat of the previously-unlinted
directories that was asked for on #809, kept away from any feature PR.


SemyonSinchenko (approved):
~LGTM

